PR #1, opened 2025-10-17
chore: adding release automation and conventional commit checking to Evaluator repo

jecortez:
This commit is meant to add [Conventional Commit](https://www.conventionalcommits.org/en/) standardization and [Release Automation](https://github.com/googleapis/release-please) to this repository to support future releases.

PR #32, opened 2026-04-08
fix: add explicit permissions to TypeScript SDK test workflow

adnanrhussain:
## Summary
  - Resolves code scanning alerts [#1](https://github.com/learning-commons-org/evaluators/security/code-scanning/1) – [#5](https://github.com/learning-commons-org/evaluators/security/code-scanning/5) - workflow was relying on default repository permissions instead of declaring explicit            
  least-privilege permissions
  - Rather than granting `actions: write` (which `styfle/cancel-workflow-action` requires), we replaced it with native `concurrency` — keeping the final permission scope to `contents: read` only

### DIFF
--- a/.github/workflows/test-sdk-typescript.yml
+++ b/.github/workflows/test-sdk-typescript.yml
@@ -16,7 +16,10 @@ on:
 
 permissions:
   contents: read
-  actions: write
+
+concurrency:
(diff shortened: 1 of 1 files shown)


georgemelvin (approved):
Looks good:
- Removes job canceling Action
- Includes native pattern to handle canceling in-progress/previous jobs aligned with PoLP

PR #34, opened 2026-04-22
feat(ts-sdk): add modelOverride option to all evaluators

adnanrhussain:
## Summary

- Adds `modelOverride: { provider, model }` to `BaseEvaluatorConfig`, letting callers swap the LLM provider and model for any evaluator without touching evaluation logic
- Adds Anthropic as a supported provider (`Provider.Anthropic`, `anthropicApiKey`)
- When `modelOverride` is set, only the API key for the override provider is required; default key validation is bypassed
- Refactors provider label tracking: `LLMProvider` interface now has a `readonly label` property computed at construction time, eliminating the per-evaluator `modelLabel` field and `getModelLabel` helper
- Telemetry: `model_override: boolean` flag when override is active; `provider` field reflects actual model used in all evaluators including `TextComplexityEvaluator`

## Test plan

- [x] Instantiate any evaluator with `modelOverride` set — confirm only the override provider's key is required, default keys are not
- [x] Check `result.metadata.model` reflects the override provider and model (e.g. `"anthropic:claude-sonnet-4-6"`)
- [x] Instantiate with `modelOverride` and a missing override key — confirm `ConfigurationError` is thrown
- [x] Run with telemetry enabled and override active — confirm `model_override: true` appears in the event and `provider` reflects the override model

### DIFF
--- a/sdks/typescript/src/providers/ai-sdk-provider.ts
+++ b/sdks/typescript/src/providers/ai-sdk-provider.ts
@@ -21,12 +21,15 @@ const DEFAULT_MODELS = {
  * Supports OpenAI, Anthropic, and Google Gemini
  */
 export class VercelAIProvider implements LLMProvider {
+  readonly label: string;
+
   constructor(private config: ProviderConfig) {
     if (config.type === 'custom') {
       throw new Error(
         'VercelAIProvider does not support custom type. Use config.customProvider directly.'
       );
     }
+    this.label = `${config.type}:${config.model ?? DEFAULT_MODELS[config.type]}`;
   }
 
   /**
--- a/sdks/typescript/src/providers/ai-sdk-provider.ts
+++ b/sdks/typescript/src/providers/ai-sdk-provider.ts
@@ -8,20 +8,12 @@ import type {
 } from './base.js';
 
 /**
- * Default models for each provider based on Python implementation
- */
-const DEFAULT_MODELS = {
-  openai: 'gpt-4o',
-  anthropic: 'claude-sonnet-4-5-20250929',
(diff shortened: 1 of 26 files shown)


sdks/typescript/src/providers/ai-sdk-provider.ts
```diff
@@ -21,12 +21,15 @@ const DEFAULT_MODELS = {
...
 export class VercelAIProvider implements LLMProvider {
+  readonly label: string;
+
   constructor(private config: ProviderConfig) {
     if (config.type === 'custom') {
       throw new Error(
         'VercelAIProvider does not support custom type. Use config.customProvider directly.'
       );
     }
+    this.label = `${config.type}:${config.model ?? DEFAULT_MODELS[config.type]}`;
```
czi-fsisenda: non-blocking: This change is only for the label, but I'm wondering, in general, if we should have a default model. Given that pricing could be very different for different models, I wonder if we should force users to pick a model themselves

czi-fsisenda (approved):
Looks good!
One non-blocking general comment.

sdks/typescript/src/providers/ai-sdk-provider.ts
adnanrhussain: This is a good callout. Ideally if the user is overriding the model, they need to specify both the provider & model (both valid). We should not silently fallback to a default model since that would hide that from the user.

Currently this was also a fallback in the `getModel` function
`const modelId = requestModel || this.config.model || this.getDefaultModel();`

I am removing this to ensure we dont silently fallback and the provider + model configuration (internal or override) are valid

PR #37, opened 2026-04-30
feat(python-sdk): python SDK scaffold

czi-fsisenda:
### Summary
Jira: 

Scaffold for evaluators for the Python sdk
* package setup
* abstract base class for all evaluators and supporting types
* using langchain to call LLMs
* abstracts: input validation, logging, metadata, etc 
* unit tests

### Test Plan
- [x] Wrote [automated tests](https://czi.atlassian.net/wiki/x/Hbl1H)
- [ ] Manually tested my changes, and here are the details:


sdks/python/src/learning_commons_evaluators/evaluators/base.py
```diff
@@ -0,0 +1,263 @@
...
+            evaluation_metadata.status = Status.failed
+            evaluation_metadata.error_details = str(e)
+            raise
+        finally:
+            evaluation_metadata.processing_time_ms = (time.perf_counter() - start) * 1000
+            self.config.logger.info(
+                "evaluation end",
+                extra={"evaluation_metadata": evaluation_metadata},
+            )
+            # TODO: send_telemetry(evaluation_metadata)
```
adnanrhussain, line 134: P1 - I dont think Telemetry is P0 for release, but valuable before first external use. Please plan accordingly

sdks/python/src/learning_commons_evaluators/schemas/config.py
```diff
@@ -0,0 +1,159 @@
...
+    OPENAI = "openai"
+
+
+@dataclass(frozen=True)
+class PromptProviderConfig:
+    """Base type for prompt provider configuration."""
+
+    api_key: str
+    type: LlmProvider
+    base_url: str | None = (
```
adnanrhussain: P0 - Is this implemented and tested? Even if optional I would remove this until we do implement this. Feel free to add a TODO stub though

sdks/python/src/learning_commons_evaluators/schemas/config.py
```diff
@@ -0,0 +1,159 @@
...
+        None  # Optional; for OpenAI-compatible endpoints (e.g. Azure, proxy). Used only when type is OPENAI.
+    )
+
+
+@dataclass(frozen=True)
+class GooglePromptProviderConfig(PromptProviderConfig):
+    """Google (Gemini) prompt provider config. Takes an API key."""
+
+    api_key: str
```
adnanrhussain: P1 - Is this redundant since its already defined in the base class?

sdks/python/src/learning_commons_evaluators/schemas/config.py
```diff
@@ -0,0 +1,159 @@
...
+class LlmProvider(str, Enum):
+    """LLM provider identifier. Subclass of str so it compares and serializes as the provider name."""
+
+    ANTHROPIC = "anthropic"
+    GOOGLE = "google"
+    OPENAI = "openai"
+
+
+@dataclass(frozen=True)
+class PromptProviderConfig:
```
adnanrhussain, line 30: P1 - LlmProvider is already the enum name in the same file, the natural rename would be `LlmProviderConfig`. 

- PromptProviderConfig → LLMProviderConfig                                                                                                                                          
- GooglePromptProviderConfig → GoogleLLMProviderConfig
- OpenAIPromptProviderConfig → OpenAILLMProviderConfig                                                                                                                              
- AnthropicPromptProviderConfig → AnthropicLLMProviderConfig

sdks/python/src/learning_commons_evaluators/schemas/config.py
```diff
@@ -0,0 +1,159 @@
...
+from enum import Enum
+
+from pydantic import BaseModel, ConfigDict
+
+from learning_commons_evaluators.logger import Logger, get_logger
+
+# --- Prompt provider configs (for LLM calls in prompt steps) ---
+
+
+class LlmProvider(str, Enum):
```
adnanrhussain, line 20: P0 - I think PEP convention is abbreviations are uppercase -> `LLMProvider`

sdks/python/src/learning_commons_evaluators/schemas/config.py
```diff
@@ -0,0 +1,159 @@
...
+    standard ``logging.Logger`` instance.
+    Create via create_config, create_config_no_telemetry, or
+    create_config_telemetry_with_full_input, then pass the config to the evaluator constructor.
+    """
+
+    google_prompt_provider_config: GooglePromptProviderConfig | None = None
+    openai_prompt_provider_config: OpenAIPromptProviderConfig | None = None
+    anthropic_prompt_provider_config: AnthropicPromptProviderConfig | None = None
+    logger: Logger = field(default_factory=get_logger)
+    telemetry_id: str | None = None
```
adnanrhussain: P0 - `telemetry_id` implies it's an identifier for a telemetry session, but what it actually is - a key issued to a partner/organization to authenticate their telemetry requests - is better captured by `partner_key`

If we do decide to change this to something else, ideally we should have the same parameter names across TS & Python

sdks/python/src/learning_commons_evaluators/schemas/config.py
```diff
@@ -0,0 +1,159 @@
...
+    send_full_input_with_telemetry: bool = False
```
adnanrhussain: P0 - I think we need an object here, to explicitly turn off telemetry. I would suggest 

```
@dataclass(frozen=True)                                                                                                                                                             
class TelemetryConfig:                                                                                                                                                              
  enabled: bool = True                                                                                                                                                            
  record_inputs: bool = False
```

sdks/python/src/learning_commons_evaluators/schemas/config.py
```diff
@@ -0,0 +1,159 @@
...
+    anthropic_prompt_provider_config: AnthropicPromptProviderConfig | None = None,
+    logger: Logger | None = None,
+) -> EvaluatorConfig:
+    """Create evaluator config with telemetry disabled."""
+    return EvaluatorConfig(
+        google_prompt_provider_config=google_prompt_provider_config,
+        openai_prompt_provider_config=openai_prompt_provider_config,
+        anthropic_prompt_provider_config=anthropic_prompt_provider_config,
+        logger=get_logger() if logger is None else logger,
+        telemetry_id=None,
```
adnanrhussain: Not a P0 here, but shortly down the line, I expect Evals built with KG to require partnerKey. So it will be used then to lookup what datasets they have access to, not just for telemetry.

sdks/python/src/learning_commons_evaluators/schemas/common_inputs.py
```diff
@@ -0,0 +1,83 @@
...
+    many field instances.
+    """
+
+    # Narrows the abstract InputField.spec: InputSpec → TextInputSpec.
+    # value: str is inherited from InputField[str] and does not need to be redeclared.
+    spec: TextInputSpec
+
+    def validate(self) -> None:
+        """Raise :class:`~.errors.ValidationError` if the value violates the spec constraints."""
+        text_length = len(self.value)
```
adnanrhussain, line 63: P1 - We should
- Strip and validate
- Default minimum should be 0
- Default maximum??? TS doesnt have it either, so lets decide on it

For errors, we should try and keep the error messages consistent b/w the too. I would start with whats currently in TS, and then doing an internal review and updating the errors messages where we need to across both, and updating the shared spec. For now

Whitespace-only input → `ValidationError("Text cannot be empty or contain only whitespace")`                                                                       
Length < min chars (trimmed) → `ValidationError("Text is too short. Minimum length is X characters...")`                                                           
Length > max chars (trimmed) → `ValidationError("Text is too long. Maximum length is X characters...")`

sdks/python/src/learning_commons_evaluators/schemas/errors.py
```diff
@@ -0,0 +1,140 @@
...
+
+class EvaluatorTimeoutError(APIError, EvaluatorRetryableError):
+    """Request timed out. Should be retried with caution."""
+
+    def __init__(self, message: str = "Request timed out"):
+        super().__init__(message, 408, True, "TIMEOUT_ERROR")
+        self.name = "EvaluatorTimeoutError"
+
+
+def _parse_provider_error(error: BaseException) -> tuple[str, int | None, str | None]:
```
adnanrhussain, line 91: P2 - I think OpenAI & Anthropic return a `status_code` in the response. For later, lets add a TODO to review the status_code if exists, and fall back to regex if absent

sdks/python/src/learning_commons_evaluators/schemas/errors.py
```diff
@@ -0,0 +1,140 @@
...
+        )
+    # Timeouts before generic "Connection" — many stacks use "Connection timed out"
+    if "timeout" in msg.lower() or "timed out" in msg.lower():
+        return EvaluatorTimeoutError(msg)
+    if any(
+        x in msg
+        for x in (
+            "ECONNREFUSED",
+            "ENOTFOUND",
+            "ETIMEDOUT",
```
adnanrhussain, line 131: P0 - Could you please confirm if these are valid for python? These look like NodeJS string properties on errors

sdks/python/src/learning_commons_evaluators/schemas/errors.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+class EvaluatorError(Exception):
+    """Base error class for all evaluator errors."""
+
+    def __init__(self, message: str, code: str | None = None):
+        super().__init__(message)
+        self.message = message
+        self.code = code
+        self.name = "EvaluatorError"
```
adnanrhussain, line 14: P0 - Isnt this `__name__` for python? This feels non-standard. Could you confirm?

sdks/python/src/learning_commons_evaluators/schemas/errors.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+class EvaluatorRetryableError(EvaluatorError):
```
adnanrhussain, line 17: P0 - `EvaluatorRetryableError` -> `RetryableError`

We can add the Evaluator prefix if it conflicts with a base python error class

sdks/python/src/learning_commons_evaluators/evaluators/base.py
```diff
@@ -0,0 +1,263 @@
...
+    Abstract base class for all evaluators.
+    Subclasses must set metadata, default_evaluation_settings, and implement evaluate_impl().
+    """
+
+    config: EvaluatorConfig
+    metadata: EvaluatorMetadata
+    default_evaluation_settings: SettingsT
+
+    def __init__(self, config: EvaluatorConfig) -> None:
+        self.config = config
```
adnanrhussain, line 68: P0 - We should validate the configuration and trigger `ConfigurationError` here, instead of during the evaluate call

sdks/python/src/learning_commons_evaluators/schemas/common_inputs.py
```diff
@@ -0,0 +1,83 @@
...
+    """
+
+    # Narrows the abstract InputField.spec: InputSpec → GradeInputSpec.
+    # value: int is inherited from InputField[int] and does not need to be redeclared.
+    spec: GradeInputSpec
+
+    def validate(self) -> None:
+        """Raise :class:`~.errors.ValidationError` if the grade is out of range or not in the allowed set."""
+        grade = self.value
+        if grade < 0 or grade > 12:
```
adnanrhussain, line 92: P0 - K-2 is only applicable for GLA. This would incorrectly lower the allowed grade bound for all the other evals

sdks/python/src/learning_commons_evaluators/schemas/metadata.py
```diff
@@ -0,0 +1,94 @@
...
+    alpha = "alpha"
+    beta = "beta"
+    rc = "rc"
+    ga = "ga"
+
+
+class Status(Enum):
+    processing = "processing"
+    succeeded = "succeeded"
+    failed = "failed"
```
adnanrhussain, line 24: P1 - Nice. For the outbound telemetry, lets keep them as `success` & `error`. If we need to change, lets discuss and get it consistent

sdks/python/src/learning_commons_evaluators/schemas/metadata.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+class EvaluationMetadata(BaseModel):
+    """Metadata for an evaluation run."""
+
+    model_config = {"arbitrary_types_allowed": True}
+
+    timestamp: datetime = Field(default_factory=lambda: datetime.now(timezone.utc))
+    evaluator_metadata: EvaluatorMetadata
+    evaluation_settings: Any
```
adnanrhussain, line 124: P1 - `EvaluationSettings`?

adnanrhussain (changes requested):

sdks/python/src/learning_commons_evaluators/evaluators/base.py
adnanrhussain: Recent scoping and sequencing discussion - Telemetry is P0

sdks/python/src/learning_commons_evaluators/schemas/common_inputs.py
czi-fsisenda: Todo

sdks/python/src/learning_commons_evaluators/schemas/common_inputs.py
adnanrhussain: Discussed - This is the bound validation, not the actual allowed grades for all evals. No action needed.

sdks/python/src/learning_commons_evaluators/schemas/config.py
czi-fsisenda: Todo or validate

sdks/python/src/learning_commons_evaluators/schemas/config.py
czi-fsisenda: `telemetry_patner_id`

sdks/python/src/learning_commons_evaluators/evaluators/base.py
czi-fsisenda: I'll do this in a follow up PR. Needs something like required_llm_providers to be defined. This PR has enough going on already.
Added a TODO.

sdks/python/src/learning_commons_evaluators/evaluators/base.py
```diff
@@ -0,0 +1,263 @@
...
+        if evaluation_settings is None:
+            evaluation_settings = self.default_evaluation_settings
+        start = time.perf_counter()
+        evaluation_metadata = EvaluationMetadata(
+            evaluator_metadata=self.metadata,
+            evaluation_settings=evaluation_settings,
+            input_metadata=input.input_values()
+            if self.config.send_full_input_with_telemetry
+            else input.input_metadata(),
+        )
```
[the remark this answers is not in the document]
czi-fsisenda: Always using `input_metadata`. Will switch it out for full input just for telemetry when telemetry is implemented.

sdks/python/src/learning_commons_evaluators/evaluators/base.py
czi-fsisenda: Got a TODO for it. To be added in subsequent PR before release.

sdks/python/src/learning_commons_evaluators/evaluators/base.py
```diff
@@ -0,0 +1,263 @@
...
+        ...
+
+    def execute_step(
+        self,
+        step_name: str,
+        evaluation_metadata: EvaluationMetadata,
+        implementation_function: Callable[[], ResultT],
+        *,
+        extras: dict[str, Any] | None = None,
+    ) -> ResultT:
```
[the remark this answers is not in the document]
czi-fsisenda: Implemented without unsafe casts.

sdks/python/src/learning_commons_evaluators/schemas/common_inputs.py
czi-fsisenda: Added strip `strip_whitespace` option to `TextInputSpec` defaulting to `True`.
So whitespace only string and empty string should fail minLength validation if there's a minimum set.

sdks/python/src/learning_commons_evaluators/schemas/common_inputs.py
czi-fsisenda: This is for generic `GradeInputField`. Each eval that uses this will set `allowed_grades` when it creates its input type.

sdks/python/src/learning_commons_evaluators/schemas/config.py
czi-fsisenda: I'll do this in a follow up since it's a refactor spanning lots of files. To be completed before launch.
Added TODO.

sdks/python/src/learning_commons_evaluators/schemas/config.py
czi-fsisenda: Same. I'll do this in a follow up since it's a refactor spanning lots of files. To be completed before launch.
Added TODO.

sdks/python/src/learning_commons_evaluators/schemas/config.py
czi-fsisenda: TODO

sdks/python/src/learning_commons_evaluators/schemas/config.py
czi-fsisenda: Fixed!

sdks/python/src/learning_commons_evaluators/schemas/config.py
czi-fsisenda: `telemetry_partner_id`

sdks/python/src/learning_commons_evaluators/schemas/config.py
czi-fsisenda: I'm taking presence of the `telemetry_id` as intent that the user wants telemetry enabled. It saves us from having the ambiguous intent where a `telemetry_id` is set, but telemetry is turned off.
Created TelemetryConfig object.

sdks/python/src/learning_commons_evaluators/schemas/config.py
czi-fsisenda: I think we should separate `partner_key` from `telemetry_partner_id` so that it's possible to use KG and an anonymous `telemery_partner_id`.
I think it's important that we let partners know that they can share telemetry anonymously and that they are free to change the anonymous id.

sdks/python/src/learning_commons_evaluators/schemas/metadata.py
czi-fsisenda: Since these are used for logging too, I'll keep them as is here and translate when sending telemetry.
Need to do telemetry translation anyway. If we stop doing translation and still prefer to change these, I can update them then.

sdks/python/src/learning_commons_evaluators/schemas/errors.py
czi-fsisenda: Correct as written according to AI, but unnecessary. I'll reorganize all of this file in a separate PR.
Added TODO. Should be P0 to avoid mixing up partners.

sdks/python/src/learning_commons_evaluators/schemas/errors.py
czi-fsisenda: Will do as part of errors reorganization. Added TODO.

sdks/python/src/learning_commons_evaluators/schemas/errors.py
czi-fsisenda: Added TODO

sdks/python/src/learning_commons_evaluators/schemas/errors.py
czi-fsisenda: Added to TODO. Address with file reorg.

sdks/python/src/learning_commons_evaluators/schemas/metadata.py
czi-fsisenda: I think this was due to Pydantic behavior where it'll prefer EvaluationSettings class over more specific class if used here. Config validation might change this.

sdks/python/Makefile
```diff
@@ -0,0 +1,51 @@
+VENV    := ../../.venv
+PYTHON  := $(VENV)/bin/python
+PYTEST  := $(VENV)/bin/pytest
+RUFF    := $(VENV)/bin/ruff
+MYPY    := $(VENV)/bin/mypy
```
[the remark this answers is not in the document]
czi-fsisenda: Updated

sdks/python/README.md
```diff
@@ -0,0 +1,334 @@
...
+baked into pre-generated Python modules at build time.  After editing any evaluator
+settings TOML under `sdks/settings/`, regenerate those modules from `sdks/python/`:
+
+```bash
+make build   # regenerates _generated_*_settings.py and syncs contracts.toml
+```
+
+Commit both the updated TOML and the regenerated `.py` files together.  CI runs
+`make check-build` to catch any drift between the TOML source and the generated files.
+
```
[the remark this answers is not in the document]
czi-fsisenda: These will be implemented before release

sdks/python/README.md
```diff
@@ -0,0 +1,334 @@
...
+# Create evaluator and run evaluation
+evaluator = ConventionalityEvaluator(config)
+result = evaluator.evaluate(
+    ConventionalityEvaluationInput(text="The cat's out of the bag now.", grade=5)
+)
+
+print(result.answer.score)        # e.g. "moderately_complex"
+print(result.answer.label)        # e.g. "Moderately complex"
+print(result.explanation.summary) # Reasoning for the score
+```
```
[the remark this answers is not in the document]
czi-fsisenda: This will be fixed before release

sdks/python/src/learning_commons_evaluators/settings/load_settings.py
```diff
@@ -0,0 +1,257 @@
...
+    *Keeping settings in sync* section of the README.
+    """
+    env = os.environ.get("EVALUATORS_SETTINGS_DIR")
+    if env:
+        return Path(env)
+    # importlib.resources.files() returns a Traversable; converting to Path
+    # works for both editable installs (a real directory) and zip/wheel installs
+    # (where Python 3.9+ extracts to a temp dir automatically).
+    pkg = importlib.resources.files("learning_commons_evaluators.settings")
+    return Path(str(pkg))
```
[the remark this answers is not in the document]
czi-fsisenda: Updated

sdks/python/src/learning_commons_evaluators/settings/load_settings.py
```diff
@@ -0,0 +1,257 @@
...
+        f"Invalid provider type '{value}' in [{step_name}]; expected one of: google, openai, anthropic."
+    )
+
+
+def _parse_prompt_settings_step(pm: dict, step_name: str) -> PromptSettings:
+    """Build PromptSettings from a TOML prompt_settings_* subsection. Raises ConfigurationError if a required field is missing."""
+    pt = pm.get("type") or pm.get("provider_type")
+    _require("type", pt, step_name)
+    if not isinstance(pt, str):
+        raise ConfigurationError(f"Field 'type' in [{step_name}] must be a string.")
```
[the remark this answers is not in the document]
czi-fsisenda: Only accepting provider_type now

adnanrhussain (approved):
lgtm. thank you for iterating

sdks/python/src/learning_commons_evaluators/evaluators/base.py
```diff
@@ -0,0 +1,286 @@
...
+            ConfigurationError: No provider config for the required LLM provider.
+            APIError (or subclasses): The LLM API call failed.
+
+        On failure, :attr:`~learning_commons_evaluators.schemas.metadata.EvaluationMetadata.status`
+        and ``error_details`` are set on the in-memory metadata object and included in the
+        evaluation end log record; no result object is returned because this method re-raises.
+        """
+        if evaluation_settings is None:
+            evaluation_settings = self.default_evaluation_settings
+        start = time.perf_counter()
```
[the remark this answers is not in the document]
czi-fsisenda: Addressed. Deep copying.

sdks/python/tests/schemas/test_input_specs.py
```diff
@@ -0,0 +1,91 @@
...
+
+        try:
+            _populate_input_spec_registry()
+            assert "_TestInputField" in INPUT_SPEC_REGISTRY
+            assert INPUT_SPEC_REGISTRY["_TestInputField"] is _TestInputSpec
+        finally:
+            # Restore the registry to the canonical state so other tests aren't affected.
+            INPUT_SPEC_REGISTRY.pop("_TestInputField", None)
+            _populate_input_spec_registry()
+
```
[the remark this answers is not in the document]
czi-fsisenda: Fixed. Restore from canonical.

sdks/python/src/learning_commons_evaluators/settings/load_settings.py
```diff
@@ -0,0 +1,192 @@
...
+
+
+def _parse_prompts(data: dict) -> dict[str, str]:
+    """Build prompts dict from TOML [prompts] section. Optional; any string values are included."""
+    prompts_section = data.get("prompts")
+    if prompts_section is None or not isinstance(prompts_section, dict):
+        return {}
+    return {
+        k: _normalize_prompt_whitespace(v if isinstance(v, str) else str(v))
+        for k, v in prompts_section.items()
```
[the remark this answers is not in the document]
czi-fsisenda: Updated

sdks/python/src/learning_commons_evaluators/logger.py
```diff
@@ -0,0 +1,98 @@
...
+        logger.addHandler(stream_handler)
+
+    return logger
+
+
+def create_silent_logger() -> logging.Logger:
+    """Logger that discards all records (no propagation). Use for ``logger=`` opt-out."""
+    logger = logging.getLogger(f"{SDK_LOGGER_NAME}.silent")
+    if not logger.handlers:
+        logger.addHandler(logging.NullHandler())
```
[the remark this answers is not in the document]
czi-fsisenda: Updated

sdks/python/src/learning_commons_evaluators/schemas/evaluator.py
```diff
@@ -0,0 +1,219 @@
...
+        errors: list[ValidationError] = []
+        for name in type(self).model_fields:
+            field_val = getattr(self, name)
+            if isinstance(field_val, InputField):
+                try:
+                    field_val.validate()
+                except ValidationError as e:
+                    errors.append(e)
+        if errors:
+            raise ValidationError(f"Validation errors: {errors}")
```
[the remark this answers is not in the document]
czi-fsisenda: Updated

PR #57, opened 2026-05-05
feat(ts-sdk): Add Purpose evaluator

adnanrhussain:
## Summary

This PR implements the Purpose evaluator to the TypeScript SDK

## Documentation
See `sdks/typescript/README.md`

## Testing
- [ ] CI workflow should build successfully

### DIFF
--- a/sdks/typescript/README.md
+++ b/sdks/typescript/README.md
@@ -379,7 +379,7 @@ await evaluator.evaluate(text: string, grade: string)
 **Returns:**
 ```typescript
 {
-  score: 'slightly_complex' | 'moderately_complex' | 'very_complex' | 'exceedingly_complex' | 'more_context_needed';
+  score: 'Slightly complex' | 'Moderately complex' | 'Very complex' | 'Exceedingly complex' | 'More context needed';
   reasoning: string;
   metadata: {
     model: string;
@@ -407,7 +407,7 @@ await evaluator.evaluate(text: string, grade: string)
 }
 ```
 
-> **Note:** The `'more_context_needed'` score is used for cases where the text alone is insufficient to determine complexity.
+> **Note:** The `'More context needed'` score is used for cases where the text alone is insufficient to determine complexity.
 
 **Example:**
 ```typescript
@@ -421,7 +421,7 @@ const result = await evaluator.evaluate(
   "The author argues that renewable energy is the only viable solution to climate change.",
   "9"
 );
-console.log(result.score);          // "moderately_complex"
+console.log(result.score);          // "Moderately complex"
 console.log(result.reasoning);
 console.log(result._internal.details.adjustment_and_scaffolding);
 ```
--- a/sdks/typescript/README.md
+++ b/sdks/typescript/README.md
@@ -352,6 +352,82 @@ await evaluator.evaluate(text: string)
 
 ---
 
+### 7. Purpose Evaluator
+
+Evaluates the Purpose dimension of qualitative text complexity — how explicitly the text's purpose is stated versus implied, and how that affects comprehension demands for the target grade level.
+
+**Supported Grades:** 3-12
+
+**Uses:** Google Gemini 3 Flash Preview
+
+**Constructor:**
+```typescript
+const evaluator = new PurposeEvaluator({
+  googleApiKey?: string;  // Google API key (required by this evaluator)
+  maxRetries?: number;    // Optional - Max retry attempts (default: 2)
+  telemetry?: boolean | TelemetryOptions; // Optional (default: true)
+  logger?: Logger;        // Optional - Custom logger
+  logLevel?: LogLevel;    // Optional - Logging verbosity (default: WARN)
+});
+```
+
+**API:**
+```typescript
+await evaluator.evaluate(text: string, grade: string)
+```
+
+**Returns:**
+```typescript
+{
+  score: 'slightly_complex' | 'moderately_complex' | 'very_complex' | 'exceedingly_complex' | 'more_context_needed';
+  reasoning: string;
+  metadata: {
+    model: string;
+    processingTimeMs: number;
+  };
+  _internal: {
+    complexity_score: 'slightly_complex' | 'moderately_complex' | 'very_complex' | 'exceedingly_complex' | 'more_context_needed';
+    reasoning: string;
+    details: {
+      detailed_summary: Array<{
+        factor: string;
+        description: string;
+        effect_on_complexity_dimension: string;
+      }>;
+      adjustment_and_scaffolding: Array<{
+        scaffolding_need: string;
+        suggestion: string;
+      }>;
+      recommended_use_cases: Array<{
+        opportunity: string;
+        suggestion: string;
+      }>;
+    };
+  };
+}
+```
+
+> **Note:** The `'more_context_needed'` score is used for cases where the text alone is insufficient to determine complexity.
+
+**Example:**
+```typescript
+import { PurposeEvaluator } from '@learning-commons/evaluators';
+
+const evaluator = new PurposeEvaluator({
+  googleApiKey: process.env.GOOGLE_API_KEY,
+});
+
+const result = await evaluator.evaluate(
+  "The author argues that renewable energy is the only viable solution to climate change.",
+  "9"
+);
+console.log(result.score);          // "moderately_complex"
+console.log(result.reasoning);
+console.log(result._internal.details.adjustment_and_scaffolding);
+```
+
+---
+
 ## Batch CSV Evaluation
 
 For evaluating many texts at once, the SDK ships a CLI tool that reads a CSV file, runs all evaluators in a group, and produces CSV and HTML reports.
@@ -474,6 +550,7 @@ interface BaseEvaluatorConfig {
 - **Conventionality**: Requires `googleApiKey` only
 - **Text Complexity**: Requires both `googleApiKey` and `openaiApiKey`
 - **Grade Level Appropriateness**: Requires `googleApiKey` only
+- **Purpose**: Requires `googleApiKey` only
 
 ---
 
--- a/sdks/typescript/src/batch/evaluator.ts
+++ b/sdks/typescript/src/batch/evaluator.ts
@@ -5,6 +5,7 @@ import {
   GradeLevelAppropriatenessEvaluator,
   SmkEvaluator,
   ConventionalityEvaluator,
+  PurposeEvaluator,
 } from '../evaluators/index.js';
 import type { BaseEvaluatorConfig } from '../evaluators/base.js';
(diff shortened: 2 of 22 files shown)


evals/prompts/purpose/input_schema.json
czi-fsisenda, line 19: [P1] As the different implementations converge, I think set is a better choice. We wouldn't be able to represent something like, 2-3, 4-6.
Not likely that we have such restrictions, but if there's a simple solution for more flexibility, we should take it.

czi-fsisenda, line 13: [P1] We should set some maximum just to be practical. Eventually, I'd like UI to pick up this value from this config.

sdks/typescript/src/batch/evaluator.ts
```diff
@@ -33,23 +34,28 @@ const EVALUATOR_MAP = new Map<string, EvaluatorConstructor>([
...
+  [PurposeEvaluator.metadata.id, PurposeEvaluator],
 ]);
 
 /**
  * Evaluator groups available for batch processing.
  * Each group runs a fixed set of evaluators and maps to a specific HTML report format.
  */
+// Note: this batch group is a superset of TextComplexityEvaluator — it also includes
+// GradeLevelAppropriateness and Purpose (early access), which the programmatic evaluator
+// does not expose. This is intentional: the batch surface evolves independently.
```
czi-fsisenda, line 46: [P1] Should we formalize that TextComplexity is an evaluation group and not an evaluator?

sdks/typescript/src/evaluators/purpose.ts
```diff
@@ -0,0 +1,197 @@
...
+// Step ID convention: "evaluate_{slug}" where slug is the last segment of evaluator.id.
+const STEP_ID = `evaluate_${CONFIG.evaluator.id.split('.').pop()}`;
+const _step = CONFIG.steps.find(s => s.id === STEP_ID);
+if (!_step) throw new Error(`Step "${STEP_ID}" not found in purpose config.json`);
+const STEP = _step;
+
+// Grade range from input_schema — needed for static metadata, so defined at module level.
+const GRADE_MIN = INPUT_SCHEMA.properties.grade_level.minimum;
+const GRADE_MAX = INPUT_SCHEMA.properties.grade_level.maximum;
+const SUPPORTED_GRADES = Array.from({ length: GRADE_MAX - GRADE_MIN + 1 }, (_, i) => String(GRADE_MIN + i));
```
czi-fsisenda, line 22: [P1] We shouldn't need to do this translation.

sdks/typescript/README.md
```diff
@@ -352,6 +352,82 @@ await evaluator.evaluate(text: string)
...
+
+**API:**
+```typescript
+await evaluator.evaluate(text: string, grade: string)
+```
+
+**Returns:**
+```typescript
+{
+  score: 'slightly_complex' | 'moderately_complex' | 'very_complex' | 'exceedingly_complex' | 'more_context_needed';
```
gary-mu: The casing changes with "_" here. eg. line 67:  we have "'slightly complex' | 'moderately complex' | 'very complex' | 'exceedingly complex'"

Wonder if we should make it consistent

sdks/typescript/src/evaluators/purpose.ts
```diff
@@ -0,0 +1,197 @@
...
+    return runPreprocessingStep(text, fkStep.implementation.typescript);
+  }
+
+  private provider: LLMProvider;
+
+  constructor(config: BaseEvaluatorConfig) {
+    super(config);
+
+    const providerType = STEP.model.provider as ProviderConfig['type'];
+    const apiKey = providerType === Providers.google ? config.googleApiKey : config.openaiApiKey;
```
gary-mu: Still learning my way around the TS side, so checking my read on this:
The code looks binary — google → Google key, anything else → OpenAI key. 

But Providers lists four options (google, openai, anthropic, custom, [‎sdks/typescript/src/providers/base.ts](https://github.com/learning-commons-org/evaluators/pull/57/changes#diff-db1f94ce2c6f5c063468393d453a69f10fb066907c6ce170644e0fb4d65616b1)). 

If a future config switched this to "anthropic", or someone use the SDK and change to "anthropic", would it silently grab the OpenAI key and hand it to an Anthropic client?

Also curious about the as ProviderConfig['type'] cast above — is that telling the compiler to trust the string without checking, which is why TS isn't flagging the missing branches? 
Def just thinking out loud from someone who has no experience with TS 🙂

gary-mu (commented):
Added some comments

sdks/typescript/src/evaluators/purpose.ts
czi-fsisenda: [P0] +1

sdks/typescript/src/evaluators/purpose.ts
```diff
@@ -0,0 +1,197 @@
...
+          input_tokens: response.usage.inputTokens,
+          output_tokens: response.usage.outputTokens,
+        },
+      });
+
+      const latencyMs = Date.now() - startTime;
+      const tokenUsage = {
+        input_tokens: response.usage.inputTokens,
+        output_tokens: response.usage.outputTokens,
+      };
```
czi-fsisenda: [P0] If you declare token usage before pushing stage details, you can use the reference and we can have 1 declaration instead of 2. Less chance of future bugs if we avoid having 2 objects that should be the same living as different objects.

sdks/typescript/src/evaluators/purpose.ts
```diff
@@ -0,0 +1,197 @@
...
+        tokenUsage,
+        metadata: { stage_details: stageDetails },
+        inputText: text,
+      }).catch(() => undefined);
+
+      this.logger.info('Purpose evaluation completed successfully', {
+        evaluator: PurposeEvaluator.metadata.id,
+        operation: 'evaluate',
+        grade: gradeNum,
+        score: result.score,
```
czi-fsisenda, line 140: [P1] should we log the whole result?

sdks/typescript/src/features/preprocessing.ts
```diff
@@ -0,0 +1,82 @@
...
+// Public API
+// ---------------------------------------------------------------------------
+
+/**
+ * Runs a single config-declared preprocessing step against `text`.
+ * Pass the language-specific implementation block (e.g. config.implementation.typescript).
+ *
+ * Throws if the library or post_transform type is not registered above.
+ */
+export function runPreprocessingStep(text: string, impl: PreprocessingImplementation): number {
```
czi-fsisenda, line 59: [P1] Feels like this should be using generics if we're planning to use it generally on different types of preprocessing.

sdks/typescript/src/evaluators/purpose.ts
```diff
@@ -0,0 +1,197 @@
...
+
+  private async callLLM(
+    inputs: Record<string, string>,
+  ): Promise<{ data: PurposeInternal; usage: { inputTokens: number; outputTokens: number }; latencyMs: number }> {
+    const response = await this.provider.generateStructured({
+      messages: [
+        { role: 'system', content: getSystemPrompt(inputs) },
+        { role: 'user', content: getUserPrompt(inputs) },
+      ],
+      schema: PurposeOutputSchema,
```
czi-fsisenda, line 198: [P1] For Purpose we have an input and output schema. LLM output happens to use output schema. But if we have multiple LLM calls, we could have intermediate processing schemas. In such cases, would we have multiple types in outputs_schema.json or multiple files or something else?

czi-fsisenda (changes requested):
Looks good! Mostly P1s. 2 P0s.

sdks/typescript/README.md
adnanrhussain: Good catch. Eventually this needs to resolve to snake_case across all the evaluators. Maybe even include the sentence case label for display if the user chooses. This will be a breaking change on the contract, so I'll do this as a separate PR, along with some other major bump changes

sdks/typescript/src/evaluators/purpose.ts
adnanrhussain: Your read is exactly right. The ternary was binary. Added explicit branches for google and openai, with a ConfigurationError thrown for anything else. This will change again with the modelOverride implementation that I have and will merge on this.

evals/prompts/purpose/input_schema.json
adnanrhussain: +1. We have identified this as a TODO in the Prompt + Notebook PR.

PS - This file is not in scope for this PR. It is from the base PR that was merged to main, and the base branch reference was switched without a rebase.

adnanrhussain: +1. Lets come back to this as an improvement.

PS - This file is not in scope for this PR. It is from the base PR that was merged to main, and the base branch reference was switched without a rebase.

sdks/typescript/src/batch/evaluator.ts
adnanrhussain: Yes. I am thinking we remove the TextComplexityEvaluator wrapper itself, and just have a generic batch evaluator that allows a group to be run in a followup PR

sdks/typescript/src/evaluators/purpose.ts
adnanrhussain: 💯 . Fixed

sdks/typescript/src/evaluators/purpose.ts
adnanrhussain: The log fields here match what SMK, Conventionality, and the others log on success - score and processingTimeMs. Logging the full result at INFO level would include reasoning and `_internal` data which can be large; if that detail is needed it's available at DEBUG. Happy to align all evaluators in a follow-up, but keeping this consistent for now.

sdks/typescript/src/features/preprocessing.ts
adnanrhussain: For the long run yes. We should keep iterating the spec and the code based on new needed complexity

sdks/typescript/src/evaluators/purpose.ts
adnanrhussain: Ack. We will followup and iterate

sdks/typescript/src/evaluators/purpose.ts
adnanrhussain: Totally. Not just LLM calls, but even for API calls. We will need to iterate the config spec

adnanrhussain:
@czi-fsisenda - Thank you for the thoughtful and forward looking feedback. I agree that we will continue to iterate and improve upon the config as we expand the capabilities we support.

Addressed the P0s and responded to your other feedback. Requesting re-review

czi-fsisenda (approved):
🚀

PR #65, opened 2026-05-16
feat(python-sdk)!: error handling improvements

czi-fsisenda:
## Summary

- Renames public validation and timeout types (`InputValidationError`, `RequestTimeoutError`) so they do not collide with builtin or Pydantic names, and documents a single `retryable` signal on `EvaluatorError` for client retry logic.
- Expands provider error wrapping: sanitized SDK messages, structured fields (`status_code`, `response_body`, `request_id`, `provider`, `model`), HTTP routing (including 408 → `RequestTimeoutError`), `OutputValidationError` for bad LLM output, and `sanitize_pydantic_errors` for safe telemetry.
- `wrap_provider_error` applies narrow timeout/network **message** patterns only when no HTTP status was extracted, so 4xx responses are not reclassified from body wording alone (aligns with the documented preference for structured signals over message parsing).
- Adds `make lint-fix` and broad README + unit test coverage for the new hierarchy.

## Test plan

- [x] Unit tests (`tests/schemas/test_errors.py`, `tests/evaluators/test_base.py`, schema and vocabulary tests updated for the new types)
- [x] Verified `TestWrapProviderError` including new cases for structured status vs. message-pattern precedence


sdks/python/README.md
```diff
@@ -302,32 +302,96 @@ config = create_config_no_telemetry(logger=create_silent_logger())
...
+log = logging.getLogger(__name__)
 try:
     result = evaluator.evaluate_sync(input)
-except ConfigurationError as e:
-    print(f"Config issue: {e}")
-except ValidationError as e:
-    print(f"Invalid input: {e}")
 except RateLimitError as e:
-    print(f"Rate limited, retry after {e.retry_after}ms")
+    time.sleep(e.retry_after or 30)  # seconds
```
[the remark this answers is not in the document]
czi-fsisenda: Fixed

sdks/python/src/learning_commons_evaluators/schemas/errors.py
```diff
@@ -1,143 +1,468 @@
...
+        "model": model,
+        "response_body": response_body,
+        "request_id": request_id,
+    }
+
+    if _is_timeout(error):
+        return RequestTimeoutError(status_code=status_code, **common)
+
+    if _is_network(error):
+        return NetworkError(**common)
```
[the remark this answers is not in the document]
czi-fsisenda: Fixed and unit tested

sdks/python/README.md
```diff
@@ -302,32 +302,98 @@ config = create_config_no_telemetry(logger=create_silent_logger())
...
+        delay = e.retry_after if isinstance(e, RateLimitError) and e.retry_after else 2 ** attempt
+        time.sleep(delay)  # retry_after is in seconds
+```
+
+### Sanitization and debugging context
+
+Error **messages** (the value returned by `str(err)`) are short and controlled. Raw provider strings — which may contain prompt echoes, user text, or fragments of API keys — are **not** interpolated into the SDK exception's message. Structured detail lives on attributes instead:
+
+- `status_code` on `APIError` — HTTP status from the provider, when one was returned. Populated from the provider exception's `.status_code` or `.response.status_code`/`.response.status` attribute when present (preferred over message regex).
+- `retry_after` on `RateLimitError` — suggested delay before retry, **in seconds**, or `None` if the provider didn't return a `Retry-After` header.
```
adnanrhussain, line 352: Note: Following the retry-after pattern, it may be helpful to adjust the TS implementation to use seconds also instead of ms. No action needed, just a callout

sdks/python/src/learning_commons_evaluators/schemas/errors.py
```diff
@@ -1,143 +1,592 @@
...
+        self,
+        message: str = "Request timed out",
+        *,
+        status_code: int | None = 408,
+        **kwargs: Any,
+    ) -> None:
+        super().__init__(message, status_code=status_code, **kwargs)
+
+
+class OutputValidationError(APIError):
```
adnanrhussain, line 222: Note: We should add this to the TS implementation also. No action needed, just calling it out

sdks/python/src/learning_commons_evaluators/schemas/common_inputs.py
```diff
@@ -59,13 +59,13 @@ def _strip_value_if_spec_requests(cls, data: Any) -> Any:
         return data
 
     def validate(self) -> None:
-        """Raise :class:`~.errors.ValidationError` if the value violates the spec constraints."""
+        """Raise :class:`~.errors.InputValidationError` if the value violates the spec constraints."""
         text_length = len(self.value)
         min_len = self.spec.min_text_length or 0
```
adnanrhussain, line 64: P2 - Not from this PR, but we may need to fallback to a default minimum > 0. Vocab's settings.toml does not specify this, and an empty input is just wasteful. I am not sure what the output behavior would be though here, an OutputValidationError, or garbage results?

adnanrhussain (approved):
LGTM. Info comments and a P2 for a followup

sdks/python/src/learning_commons_evaluators/schemas/common_inputs.py
czi-fsisenda: I'm hoping we move towards a standard text complexity input so individual evals don't need to specify this.
But, yes, we should have > 0.
I think we should set the value rather than some arbitrary default. 1 is about as useless as 0.

PR #77, opened 2026-05-22
feat(ts-sdk): add bypassRowLimit option for batch evaluator

### LINKED ISSUE #75: [FEEDBACK]: IMHO, 50 is too small a limit for the evaluators-batch tool
### What is your feedback about?  
*Tell us which tool, feature, or part of the experience you're referring to.*
The `evaluators-batch` tool ([here](https://github.com/learning-commons-org/evaluators/blob/main/sdks/typescript/src/batch/README.md)) has a hard limit of 50 rows. I find this very limiting. I would want to be able to run this on a larger batch and step away from my desk. Or run it on a cloud machine and have it run overnight.

### Please describe your feedback.  
*What worked well, didn’t work, or could be improved? Be as specific as possible.*

I love the batch CLI tool. But as per above, a higher limit would be nice.

### Why is this feedback important to you?  
*How did this impact your workflow, goals, or expectations?*

Would simply make it easier to run large data sets through the tool. 50 rows is way too small for my needs (evaluating large numbers of AI responses being sent to students). 500 or 1000 would be more reasonable.

### Additional context  
*Include any examples, links, screenshots, or notes that could help us better understand your feedback.*

I'm referring to the code at https://github.com/learning-commons-org/evaluators/blob/8665aaa576bcabafa53f3ee6a664c0e7a0d2e02c/sdks/typescript/src/batch/evaluator.ts#L62.

Alternatively, maybe allow an input flag that overrides this?

adnanrhussain:
## Summary
- Adds `bypassRowLimit` to `BatchConfig` and `--bypass-row-limit` CLI flag, letting users opt out of the per-group 50-row hard limit when they accept the longer-runtime / provider-throttling risk.
- Default behavior is unchanged — limit is still enforced unless the flag is explicitly set. Error message and CLI suggestion now point users at the escape hatch.
- Resolves #75

## Test plan
- [x] `npm run build` clean
- [x] `npm test` passes
- [x] `npm run lint` — 0 errors
- [x] Manual CLI smoke test with 51-row CSV: error path lists `--bypass-row-limit`; `--bypass-row-limit` shows warning and proceeds to key prompts
- [x] Programmatic verification covered by `bypassRowLimit: true skips the row limit check` unit test

### DIFF
--- a/sdks/typescript/src/batch/README.md
+++ b/sdks/typescript/src/batch/README.md
@@ -61,7 +61,7 @@ Any additional columns beyond `text` and `grade` are preserved as-is in the outp
 
 The batch evaluator runs a fixed group of evaluators together. The current available group is:
 
-- **text-complexity**: Runs grade-level appropriateness, subject matter knowledge, vocabulary complexity, sentence structure, conventionality, and purpose evaluators together (requires both Google and OpenAI API keys). Maximum 50 input rows. If you exceed the limit, the CLI will exit with an error and suggest splitting into smaller batches.
+- **text-complexity**: Runs grade-level appropriateness, subject matter knowledge, vocabulary complexity, sentence structure, conventionality, and purpose evaluators together (requires both Google and OpenAI API keys). Maximum 50 input rows. If you exceed the limit, the CLI will exit with an error and suggest splitting into smaller batches. The limit can be bypassed with `--bypass-row-limit` — see [CLI Flags](#cli-flags) below.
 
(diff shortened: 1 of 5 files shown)


georgemelvin (approved):
Looks good:
- includes optional bypass row limit variable to SDK interface to support user defined batch sizes
- includesoptional bypass row limit variable to CLI to support user defined batch sizes
-  updates testing
- defers telemetry to capture usage to later

PR #81, opened 2026-05-22
chore(ts-sdk): migrate to eslint 10 and @typescript-eslint v8

adnanrhussain:
## Summary

Dependabot opened three PRs that individually can't merge because they share a peer-dep conflict (`@typescript-eslint/eslint-plugin@6` only supports `eslint@^7 || ^8`, blocking the ESLint 10 bump):

  - #69
  - #68
  - #67

ESLint 10 also removes the legacy `.eslintrc` config format entirely, so the bump requires a flat-config migration as well. This PR does all four pieces in one coordinated change so the upgrade can actually land.

(Older variants of the same bumps — #45, #54, #55, #56 — are superseded and will close out automatically.)

#### Changes

  - Bumped `eslint` 8.56 → 10.4, `@typescript-eslint/{eslint-plugin,parser}` 6.19 → 8.59 (latest stable for all)
  - Added `@eslint/js` and `globals` (required by flat config)
  - Replaced `.eslintrc.cjs` with `eslint.config.js` (flat config)
  - Updated `lint` script: `eslint src tests --ext .ts` → `eslint "src/**/*.ts" "tests/**/*.ts"` (`--ext` removed in ESLint 9+)
  - Disabled the base `no-redeclare` rule (doesn't understand TS type/value pattern used in zod schemas)
  - Removed two stale `eslint-disable` directives that v10 correctly reports as unused

  ## Testing

  All gates green locally:
  - [x] `npm run lint`
  - [x] `npm run typecheck`
  - [x] `npm run build`
  - [x] `npm run test:unit`
  - [x] `npm run test:coverage`


czi-fsisenda (approved):
LGTM! Copilot suggestion seems reasonable to address, but it's a nit.

PR #95, opened 2026-06-09
feat(batch): CLI enhancements and model override support

adnanrhussain:
## Summary

**CLI enhancements**
- `--help` / `--version` — documents all flags; version pulled from the SDK
- Positional CSV arg — `evaluators-batch input.csv` skips the file path prompt; falls back to interactive if omitted
- `--google-api-key`, `--openai-api-key`, `--anthropic-api-key` — skip the interactive prompt when provided via flag or env var (`GOOGLE_API_KEY`, `OPENAI_API_KEY`, `ANTHROPIC_API_KEY`)
- `--model <provider:model>` — global model override for all evaluators (e.g. `--model anthropic:claude-opus-4-8`); validated early before any prompts; when set, only the override provider's key is required
- `--output-dir <path>` — skips the output directory prompt
- Row limit error prints a copy-pasteable re-run command with `--bypass-row-limit` appended; API key values are redacted
- Confirm prompt always shown — intentional safety checkpoint before expensive runs
- After completion, the full path to the HTML report is printed so the user can open it directly

**Security**
- Auto browser-open removed — eliminates the subprocess/CodeQL concern; report path is printed to terminal instead

**Type changes**
- `BatchTask` gains `originalRow` (was always added inline as an ad-hoc intersection — now part of the type)
- `BatchConfig` gains `anthropicApiKey` and `modelOverride`
- `BatchEvaluator` default `telemetry` changed to `true` — consistent with the SDK default

**Testability**
- Pure CLI logic (`parseArgs`, `parseModelOverride`, `requiredProviders`) extracted to `cli-args.ts` so it can be unit tested without triggering `main()`

## Test plan

- [x] `npx tsc --noEmit` passes (no type errors)
- [x] `vitest run tests/unit/batch` — 96 tests pass (up from 49 on main)
- [x] `evaluators-batch --help` prints full usage and exits
- [x] `evaluators-batch --version` prints SDK version and exits
- [x] `evaluators-batch input.csv` skips the file path prompt
- [x] `GOOGLE_API_KEY=xxx OPENAI_API_KEY=xxx evaluators-batch` skips both key prompts
- [x] `--model anthropic:claude-opus-4-8` with `ANTHROPIC_API_KEY` set skips Google + OpenAI prompts
- [x] `--model bad-provider:foo` exits with a clear error before any prompts appear
- [x] Row limit exceeded prints re-run command with `--bypass-row-limit` appended; API key values are redacted
- [x] `--output-dir ./out` with real API keys writes `results.csv` and `results.html` to the specified directory and prints the full HTML path

### DIFF
--- a/sdks/typescript/src/batch/cli-args.ts
+++ b/sdks/typescript/src/batch/cli-args.ts
@@ -31,23 +31,28 @@ export function parseArgs(argv: string[] = process.argv.slice(2)): CliArgs {
       result.noTelemetry = true;
     } else if (arg === '--bypass-row-limit') {
       result.bypassRowLimit = true;
-    } else if (arg === '--concurrency' && argv[i + 1]) {
+    } else if (arg === '--concurrency' && argv[i + 1] && !argv[i + 1].startsWith('--')) {
       const v = parseInt(argv[++i], 10);
       if (!isNaN(v) && v > 0) result.concurrency = v;
-    } else if (arg === '--max-retries' && argv[i + 1]) {
+    } else if (arg === '--max-retries' && argv[i + 1] && !argv[i + 1].startsWith('--')) {
       const v = parseInt(argv[++i], 10);
       if (!isNaN(v) && v >= 0) result.maxRetries = v;
-    } else if (arg === '--google-api-key' && argv[i + 1]) {
+    } else if (arg === '--google-api-key' && argv[i + 1] && !argv[i + 1].startsWith('--')) {
       result.googleApiKey = argv[++i];
-    } else if (arg === '--openai-api-key' && argv[i + 1]) {
+    } else if (arg === '--openai-api-key' && argv[i + 1] && !argv[i + 1].startsWith('--')) {
       result.openaiApiKey = argv[++i];
-    } else if (arg === '--anthropic-api-key' && argv[i + 1]) {
+    } else if (arg === '--anthropic-api-key' && argv[i + 1] && !argv[i + 1].startsWith('--')) {
       result.anthropicApiKey = argv[++i];
-    } else if (arg === '--output-dir' && argv[i + 1]) {
+    } else if (arg === '--output-dir' && argv[i + 1] && !argv[i + 1].startsWith('--')) {
       result.outputDir = argv[++i];
-    } else if (arg === '--model' && argv[i + 1]) {
+    } else if (arg === '--model' && argv[i + 1] && !argv[i + 1].startsWith('--')) {
       result.model = argv[++i];
-    } else if (!arg.startsWith('--') && !result.csv) {
+    } else if (arg.startsWith('--')) {
+      // Unknown flag — skip its value if the next token doesn't look like a flag
+      if (argv[i + 1] && !argv[i + 1].startsWith('--')) {
+        i++;
+      }
+    } else if (!result.csv) {
       result.csv = arg;
     }
   }
--- a/sdks/typescript/src/batch/cli-args.ts
+++ b/sdks/typescript/src/batch/cli-args.ts
@@ -42,21 +42,21 @@ export function parseArgs(argv: string[] = process.argv.slice(2)): CliArgs {
       result.noTelemetry = true;
     } else if (arg === '--bypass-row-limit') {
       result.bypassRowLimit = true;
-    } else if (arg === '--concurrency' && args[i + 1] && !args[i + 1].startsWith('-')) {
+    } else if (arg === '--concurrency' && i + 1 < args.length && !args[i + 1].startsWith('-')) {
       const v = parseInt(args[++i], 10);
       if (!isNaN(v) && v > 0) result.concurrency = v;
-    } else if (arg === '--max-retries' && args[i + 1] && !args[i + 1].startsWith('-')) {
+    } else if (arg === '--max-retries' && i + 1 < args.length && !args[i + 1].startsWith('-')) {
       const v = parseInt(args[++i], 10);
       if (!isNaN(v) && v >= 0) result.maxRetries = v;
-    } else if (arg === '--google-api-key' && args[i + 1] && !args[i + 1].startsWith('-')) {
+    } else if (arg === '--google-api-key' && i + 1 < args.length && !args[i + 1].startsWith('-')) {
       result.googleApiKey = args[++i];
-    } else if (arg === '--openai-api-key' && args[i + 1] && !args[i + 1].startsWith('-')) {
+    } else if (arg === '--openai-api-key' && i + 1 < args.length && !args[i + 1].startsWith('-')) {
       result.openaiApiKey = args[++i];
-    } else if (arg === '--anthropic-api-key' && args[i + 1] && !args[i + 1].startsWith('-')) {
+    } else if (arg === '--anthropic-api-key' && i + 1 < args.length && !args[i + 1].startsWith('-')) {
       result.anthropicApiKey = args[++i];
-    } else if (arg === '--output-dir' && args[i + 1] && !args[i + 1].startsWith('-')) {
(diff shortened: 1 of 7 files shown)


sdks/typescript/src/batch/evaluator.ts
```diff
@@ -86,12 +83,12 @@ export class BatchEvaluator {
     this.config = {
       concurrency: 3,
       maxRetries: 2,
-      telemetry: false,
+      telemetry: true, // matches CLI default; opt out with { telemetry: false } or --no-telemetry
       bypassRowLimit: false,
       ...config,
```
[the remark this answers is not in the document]
adnanrhussain: Making batch eval symmetrical to the sdk, with telemetry on by default.

sdks/typescript/src/batch/cli-args.ts
```diff
@@ -0,0 +1,102 @@
...
+    } else if (arg === '--concurrency' && i + 1 < args.length && !args[i + 1].startsWith('-')) {
+      const v = parseInt(args[++i], 10);
+      if (!isNaN(v) && v > 0) result.concurrency = v;
+    } else if (arg === '--max-retries' && i + 1 < args.length && !args[i + 1].startsWith('-')) {
+      const v = parseInt(args[++i], 10);
+      if (!isNaN(v) && v >= 0) result.maxRetries = v;
+    } else if (arg === '--google-api-key' && i + 1 < args.length && !args[i + 1].startsWith('-')) {
+      result.googleApiKey = args[++i];
+    } else if (arg === '--openai-api-key' && i + 1 < args.length && !args[i + 1].startsWith('-')) {
+      result.openaiApiKey = args[++i];
```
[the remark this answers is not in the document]
adnanrhussain: This contradicts an earlier finding on this PR that correctly advised using startsWith('-') to prevent short flags like -h from being consumed as values. No real API key from Google, OpenAI, or Anthropic starts with - — this is purely theoretical. The current behaviour is intentional and consistent with the re-run hint.

sdks/typescript/src/batch/index.ts
```diff
@@ -27,3 +27,5 @@ export type {
   BatchConfig,
   BatchSummary,
 } from './types.js';
+export { Provider } from '../evaluators/base.js';
+export type { ModelOverride } from '../evaluators/base.js';
```
esthersweon:
```suggestion
export { Provider, type ModelOverride } from '../evaluators/base.js';
```

sdks/typescript/src/batch/evaluator.ts
```diff
@@ -86,12 +83,12 @@ export class BatchEvaluator {
     this.config = {
       concurrency: 3,
       maxRetries: 2,
-      telemetry: false,
+      telemetry: true, // matches CLI default; opt out with { telemetry: false } or --no-telemetry
```
esthersweon: Would remove comment about default, since it's a given (and true for the other config options).

```suggestion
      telemetry: true, // Opt out with --no-telemetry
```

sdks/typescript/src/batch/evaluator.ts
```diff
@@ -86,12 +83,12 @@ export class BatchEvaluator {
...
       bypassRowLimit: false,
       ...config,
     };
 
-    this.limit = pLimit(this.config.concurrency!);
+    this.limit = pLimit(this.config.concurrency ?? 3);
```
esthersweon, line 91: Is this necessary if we set the 3 default above?

sdks/typescript/src/batch/cli.ts
```diff
@@ -1,171 +1,272 @@
...
+    const code = (error as { code?: string }).code;
+    if (code === 'EACCES') return `No write permission for directory: ${checkDir}`;
+    if (code === 'EROFS')  return `Directory is read-only: ${checkDir}`;
+    return `Cannot write to directory: ${error instanceof Error ? error.message : String(error)}`;
+  }
+}
+
+// ---- Help / version ----
+
+function printHelp(): void {
```
esthersweon, line 60: #nit can we use one `console.log` (with newlines) for easier scanning of the output here?

sdks/typescript/src/batch/cli.ts
```diff
@@ -1,171 +1,272 @@
...
-    // (Only one group currently; when more are added this becomes a selection prompt)
+    // Step 2: Show the evaluator group that will run
     const group = getAvailableGroups()[0];
     console.log(`✓ Evaluator group: ${group.name}`);
-    console.log(`  ${group.description}\n`);
+    console.log(`  ${group.description}`);
+    if (modelOverride) {
+      console.log(`  ⚡ Model override: ${modelOverride.provider}:${modelOverride.model}`);
+    }
+    console.log();
```
esthersweon, line 202: Remove?

sdks/typescript/src/batch/cli-args.ts
```diff
@@ -0,0 +1,102 @@
+import { Provider } from '../evaluators/base.js';
+import type { ModelOverride } from '../evaluators/base.js';
+import type { EvaluatorGroup } from './types.js';
+
+export interface CliArgs {
+  csv?: string;
```
esthersweon: `pathToCsv` or `csvPath` to be more explicit?

sdks/typescript/src/batch/cli-args.ts
```diff
@@ -0,0 +1,102 @@
...
+  concurrency?: number;
+  maxRetries?: number;
+  noTelemetry?: boolean;
+  bypassRowLimit?: boolean;
+  googleApiKey?: string;
+  openaiApiKey?: string;
+  anthropicApiKey?: string;
+  outputDir?: string;
+  model?: string;
```
esthersweon: `providerModel` to be clearer about expected format?

esthersweon (approved):
Left some minor comments, but overall looks good to me.

Confirming that this would go out in a v0.7.0 release for TS?

sdks/typescript/src/batch/evaluator.ts
adnanrhussain: deliberate safety guard - if a caller explicitly passes `{ concurrency: undefined }`, the spread overwrites the default 3 with `undefined`

sdks/typescript/src/batch/cli.ts
adnanrhussain: intentional for separating the log groups

sdks/typescript/src/batch/cli-args.ts
```diff
@@ -0,0 +1,102 @@
...
+    } else if (arg === '--concurrency' && i + 1 < args.length && !args[i + 1].startsWith('-')) {
+      const v = parseInt(args[++i], 10);
+      if (!isNaN(v) && v > 0) result.concurrency = v;
+    } else if (arg === '--max-retries' && i + 1 < args.length && !args[i + 1].startsWith('-')) {
+      const v = parseInt(args[++i], 10);
+      if (!isNaN(v) && v >= 0) result.maxRetries = v;
+    } else if (arg === '--google-api-key' && i + 1 < args.length && !args[i + 1].startsWith('-')) {
+      result.googleApiKey = args[++i];
+    } else if (arg === '--openai-api-key' && i + 1 < args.length && !args[i + 1].startsWith('-')) {
+      result.openaiApiKey = args[++i];
```
czi-fsisenda, line 54: [P1] Consider reading these from env variables instead / also

czi-fsisenda (approved):
LGTM!

sdks/typescript/src/batch/cli-args.ts
adnanrhussain: https://github.com/learning-commons-org/evaluators/pull/95/changes#diff-9738637a14110bf082a6c4e1e690c7a1aaf29444b618aa0d39c0072ed2a172c6R99-R110

We do
```
...
if (fromEnv) return fromEnv;
...
```

PR #99, opened 2026-06-11
fix: enforce strict semver for python SDK

adnanrhussain:
## Summary

- Removes both `bump-patch-for-minor-pre-major` and `bump-minor-pre-major` from the `sdks/python` config so the Python SDK follows strict semver
- `feat:` → minor bump (was incorrectly producing patch bumps, e.g. GLA evaluator in #92 produced `0.1.1` instead of `0.2.0`)
- `feat!:` / `BREAKING CHANGE` → major bump (intentional path to `1.0.0`)

## After merging

Release-please will regenerate PR #88 and should recalculate the Python SDK version as `0.2.0`.

### DIFF
--- a/release-please-config.json
+++ b/release-please-config.json
@@ -58,8 +58,7 @@
             "release-type": "python",
             "changelog-path": "CHANGELOG.md",
             "component": "sdks-python",
-            "bump-minor-pre-major": true,
-            "bump-patch-for-minor-pre-major": true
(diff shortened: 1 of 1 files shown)


georgemelvin (approved):
Looks good:
- includes support for strict semver squencing

czi-fsisenda (approved):
🚀

PR #105, opened 2026-06-15
fix(typescript-sdk): prevent bundlers from requiring unused @ai-sdk providers

### LINKED ISSUE #109: [BUG]: Next.js build fails with "Module not found: Can't resolve '@ai-sdk/anthropic'" — only installed @ai-sdk/openai
### Describe the bug
We're using the evaluators SDK in a Next.js app and we only use OpenAI as our model provider, so I only installed `@ai-sdk/openai`. When I run the app in Next development, the build fails with:

```
Module not found: Can't resolve '@ai-sdk/anthropic'
Module not found: Can't resolve '@ai-sdk/google'
```

It seems like the SDK expects all of the provider packages to be present even though I'm only using one. The only ways I've found to get past it are:

- Install all three (`@ai-sdk/anthropic` and `@ai-sdk/google` as well), even though we don't use them, or
- Add `@learning-commons/evaluators` to `serverExternalPackages` in my Next config — but that doesn't really seem correct and I'm not sure it's the right thing to do.

Ideally we'd only need to install the provider we're actually using.

### List steps to reproduce the issue
1. Create a Next.js app.
2. Install `@learning-commons/evaluators` and only the OpenAI adapter (`@ai-sdk/openai`), plus `ai` and `zod`.
3. Use the SDK with an OpenAI model override:
   ```ts
   import {
     type BaseEvaluatorConfig,
     LogLevel,
     Provider,
   } from "@learning-commons/evaluators";

   const learningCommonsConfig: BaseEvaluatorConfig = {
     openaiApiKey: process.env.OPENAI_API_KEY,
     modelOverride: {
       provider: Provider.OpenAI,
       model: "gpt-4o-mini",
     },
     telemetry: false,
     logLevel: LogLevel.SILENT,
   };

   const evaluator = new VocabularyEvaluator(learningCommonsConfig);
   const result = await evaluator.evaluate(text, "8");
   ```
4. Run the app / build it with Next.
5. The build fails with `Module not found: Can't resolve '@ai-sdk/anthropic'` and `@ai-sdk/google`.

### Expected behavior
I should only need to install the provider adapter we actually use (`@ai-sdk/openai`). The app should build without requiring `@ai-sdk/anthropic` or `@ai-sdk/google`, and without having to add the package to `serverExternalPackages`.

### Environment (please complete the following):
- Node: 24.x
- Next.js: 16.2.6
- React: 19.2.6
- `@learning-commons/evaluators`: 0.6.0
- `@ai-sdk/openai`: 3.0.67
- `ai`: 6.0.193
- `zod`: 3.25.76
- Package manager: pnpm

### Additional context
_(none)_

adnanrhussain:
## Summary

Resolves #109 

Consumers only need to install the `@ai-sdk/*` provider they actually use, but bundlers (Next.js/Turbopack, webpack, Vite/Rollup) statically resolve the SDK's dynamic `import()` specifiers at build time, so a build with only `@ai-sdk/openai` installed fails with `Module not found: Can't resolve '@ai-sdk/anthropic'` (and `@ai-sdk/google'`).

This adds `webpackIgnore` / `turbopackIgnore` / `@vite-ignore` magic comments to the dynamic provider imports so bundlers leave them as runtime calls. The existing `.catch()` guard still produces a friendly "install its adapter" error if a selected provider isn't installed.

## Validation

Reproduced and verified with a Next.js 16.2.6 app (only `@ai-sdk/openai` installed):

| | Before | After |
|---|---|---|
| `next build` | ❌ `Module not found` for `@ai-sdk/anthropic` + `@ai-sdk/google` | ✅ compiles successfully |

Build, lint, and 254/254 unit tests pass.

### DIFF
--- a/sdks/typescript/src/providers/ai-sdk-provider.ts
+++ b/sdks/typescript/src/providers/ai-sdk-provider.ts
@@ -83,14 +83,22 @@ export class VercelAIProvider implements LLMProvider {
 
   /**
    * Get the configured language model.
-   * Uses dynamic imports so consumers only need to install the provider packages they use.
+   *
+   * Uses dynamic imports so consumers only need to install the provider packages
+   * they actually use. The `webpackIgnore` / `turbopackIgnore` / `@vite-ignore`
+   * magic comments stop bundlers (Next.js, webpack, Vite/Rollup, etc.) from
+   * statically resolving the uninstalled providers at build time — without them,
+   * bundling fails with "Module not found: Can't resolve '@ai-sdk/...'" even
+   * though the import is guarded at runtime.
    */
   private async getModel() {
     const apiKey = this.config.apiKey;
 
     switch (this.config.type) {
       case 'openai': {
-        const { createOpenAI } = await import('@ai-sdk/openai').catch(() => {
+        const { createOpenAI } = await import(
+          /* webpackIgnore: true */ /* turbopackIgnore: true */ /* @vite-ignore */ '@ai-sdk/openai'
+        ).catch(() => {
           throw new Error(
             'To use the OpenAI provider, install its adapter: npm install @ai-sdk/openai'
           );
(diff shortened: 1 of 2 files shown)


georgemelvin (approved):
Looks good:
- pushes import of LLM provider(s) to be dynamically imported at runtime. This resolves issue of requiring all providers to be present at buildtime.

PR #108, opened 2026-06-18
feat(ts-sdk): jurisdiction support for math standards alignment

adnanrhussain:
## Summary

- Adds `Jurisdiction` enum (52 values, matching KG API) to the KG client layer so future evaluators can reuse it
- Makes `jurisdiction` a mandatory top-level parameter on all evaluator entry points (`evaluate`, `evaluateItems`, `evaluateQuestionBank`, `evaluateByGrade`) — forces explicit choice, no silent Multi-State default
- KG client is now subject-agnostic: `academicSubject` is passed by the caller; math evaluator passes `'Mathematics'` internally via `KG_SUBJECT`
- Switches from throw-on-ambiguous to `limit=1` + take-first for standard lookups — eliminates deduplication logic and matches API design intent
- Removes `grade` from the public `evaluate()` signature and from system/user prompt templates; grade is still recorded internally when available (via `evaluateItems`/`evaluateQuestionBank`) but is `grade?: string` on `StandardAlignmentResult`
- Framework UUID lookups for non-Multi-State jurisdictions use a cached call to `/standards-frameworks`

## Test plan

- [x] `npm test` — 278 pass, 0 fail
- [x] `npm run lint` — 0 errors
- [x] `evaluate(question, statementCode, Jurisdiction.California)` passes `jurisdiction=California` and `academicSubject=Mathematics` to the KG search endpoint
- [x] `evaluateByGrade(questions, '3', Jurisdiction.California)` calls `/standards-frameworks?jurisdiction=California&academicSubject=Mathematics` to get the correct framework UUID


sdks/typescript/src/knowledge-graph/client.ts
```diff
@@ -61,20 +72,25 @@ export class KnowledgeGraphClient {
...
+    const jurisdiction = opts?.jurisdiction ?? 'Multi-State';
+    const cacheKey = `${statementCode}:${jurisdiction}`;
+    let p = this.standardInfoCache.get(cacheKey);
     if (!p) {
-      p = this.limit(() => this._fetchStandardInfo(statementCode));
-      p = p.catch((err) => { this.standardInfoCache.delete(statementCode); throw err; });
-      this.standardInfoCache.set(statementCode, p);
+      p = this.limit(() => this._fetchStandardInfo(statementCode, opts));
+      p = p.catch((err) => { this.standardInfoCache.delete(cacheKey); throw err; });
+      this.standardInfoCache.set(cacheKey, p);
```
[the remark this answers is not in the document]
adnanrhussain: No valid. For this eval, its only going to be Math. And with limit as 1, deterministically it should be the same one KG API returns and is cached

sdks/typescript/src/knowledge-graph/client.ts
```diff
@@ -61,20 +72,25 @@ export class KnowledgeGraphClient {
   private readonly apiKey: string;
   private readonly limit: ReturnType<typeof pLimit>;
 
+  // Cache key: `${statementCode}:${jurisdiction}`
```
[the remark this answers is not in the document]
adnanrhussain: The comment is accurate as written — the cache key is `${statementCode}:${jurisdiction}` and does not vary by `academicSubject` or `limit`. Those are pass-through fetch parameters but not part of the cache key, by design: all math evaluator calls use the same `academicSubject` ('Mathematics') and `limit` (1), so the `statementCode + jurisdiction` pair is sufficient for cache identity.

evals/prompts/math/standards-alignment/system.txt
```diff
@@ -1,4 +1,4 @@
-You are an expert in math Common Core State Standards for grade {grade}.
+You are an expert in math Common Core State Standards.
```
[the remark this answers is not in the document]
adnanrhussain: Updated

evals/prompts/math/standards-alignment/system.txt
```diff
@@ -1,61 +1 @@
...
-* State whether the required evidence matches. If not, name the mismatch precisely.
-
-Feedback when aligned = false:
-* Give a concrete revision that would make the item align.
-* Preserve the original context when possible, but change the task so students must demonstrate the target skill.
-
-Decision standard:
-* Prefer true when the item clearly elicits the target mathematical work.
-* Prefer false when alignment depends on speculation, related-but-different skills, or only partial overlap.
+You are an expert in K-12 math academic standards.
```
adnanrhussain, line 1: Making this consistent back w/ the source notebook

evals/prompts/math/standards-alignment/user.txt
```diff
@@ -1,9 +1,100 @@
-Evaluate the assessment question below against each of the {n} learning components listed. Return a JSON object with an "evaluations" array. Each entry must include the "lc_id" field copied exactly from the identifier shown in brackets — this is how we verify the result maps to the right learning component.
```
adnanrhussain, line 1: Making this consistent back w/ the source notebook. Minimal changes enable evaluating multiple learning components, vs a single component

sdks/typescript/src/schemas/math/standards-alignment.ts
```diff
@@ -5,7 +5,7 @@ import { z } from 'zod';
 export const LCEvaluationSchema = z.object({
   lc_id: z.string(),
   reasoning: z.string(),
-  aligned: z.boolean(),
+  answer: z.enum(['Yes', 'No']),
```
czi-fsisenda, line 8: [P2] Been trying to establish a standard output shape for evaluation results. Could it work for this?
https://github.com/learning-commons-org/evaluators/blob/70362e464506a7d7bb50f675547a8f0f89cf97b8/sdks/python/src/learning_commons_evaluators/schemas/evaluator.py#L194-L220

sdks/typescript/src/knowledge-graph/client.ts
czi-fsisenda, line 1: Generated, right?

czi-fsisenda (approved):
LGTM! 🚀

sdks/typescript/src/knowledge-graph/client.ts
adnanrhussain: the client is not generated, since I have some caching, concurrency limits, etc. But I could do a followup and break that apart, and just have the generated faw fetch and the wrappers on top

PR #119, opened 2026-06-25
chore: add repo check harness with notebook stripping

adnanrhussain:
## What

Adds a repo-wide **check harness** under `scripts/`, with its first check: stripping transient state from Jupyter notebooks. The same harness runs locally and in CI.

```bash
pip install -r scripts/requirements.txt
python scripts/check.py          # verify (what CI runs)
python scripts/check.py --fix    # auto-fix what's safe
```

## Why

Notebook `outputs` / `execution_count` / cell metadata kept landing in source control (e.g. recent eval PRs), causing noisy diffs, merge conflicts, and a risk of leaking data/keys. There was no automation catching it. This establishes a self-service pattern: **break early, with clear info, auto-fix when possible.**

## How it works

- One check per concern, two modes: `--check` (verify, non-zero exit) and `--fix` (repair in place).
- The orchestrator runs **every** check and reports **all** failures at once — no stop-at-first.
- `strip-notebooks` drives **pinned `nbstripout==0.9.1`** (`scripts/requirements.txt`) with `--keep-id` (no cell-id churn) and `--extra-keys` for the Databricks + Jupyter (`outputs_hidden`) cell metadata nbstripout keeps by default.
- CI (`.github/workflows/checks.yml`) runs the same harness on PRs/pushes and **fails only** (never auto-fixes).

## This PR also

Strips the 9 currently-dirty tracked notebooks. Verified: **0 source-content changes** (state-only); the 5 already-clean notebooks are untouched.

## Follow-ups (noted in `scripts/README.md`)

eval-config validation, naming/structure checks, delegation to the SDK suites, and a root-README/CONTRIBUTING pointer once there are more checks.

PR #120, opened 2026-06-25
chore: add eval-config validation check

adnanrhussain:
> **Stacked on #119** (`ahussain/nbstripout`). Review/merge that first; this PR's diff is against that base.

## What

Adds a second check to the harness: **`eval-configs`**, validating every `evals/**/config.json` against repo conventions.

```bash
python scripts/check.py eval-configs
```

## Rules enforced

1. referenced files exist (`$ref` schemas, message `source_path`, fixtures path)
2. config + referenced schema/fixture files are valid JSON
3. each message's declared `sha256` matches the prompt file (drift tripwire)
4. `parser.kind == "structured_output"`
5. message roles ∈ {system, user, assistant}
6. placeholder/template consistency (declared ⇄ used)

Check-only — config intent isn't safely auto-fixable (notably `sha256`, a deliberate drift tripwire).

## It immediately caught a real leftover

`intertextuality/config.json` still declared a `format_instructions` placeholder (and stale `parser` sub-fields) from the old LangChain pattern — unused since the switch to `structured_output`. Fixed here so the repo is green.

## Deferred (noted in `scripts/README.md`)

Fixture-label-in-enum validation needs a declared fixture→output-field mapping first, since some evaluators rename the field (`complexity_score` → `complexity_level`). Naming/structure conventions to follow.


scripts/checks/eval_config.py
```diff
@@ -0,0 +1,164 @@
...
+            fail("missing $schema reference")
+            return
+        schema_path = os.path.normpath(os.path.join(base, ref))
+        if not os.path.exists(schema_path):
+            fail(f"$schema not found: {ref}")
+            return
+        validator = Draft202012Validator(load_json(schema_path))
+        for err in sorted(validator.iter_errors(config), key=lambda e: list(e.path)):
+            loc = "/".join(str(p) for p in err.path) or "(root)"
+            fail(f"schema: {loc}: {err.message}")
```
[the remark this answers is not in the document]
adnanrhussain: Fixed in 8f5109c: `_check_schema` now guards the schema read and reports a violation instead of crashing.

scripts/checks/eval_config.py
```diff
@@ -0,0 +1,164 @@
...
+
+    @staticmethod
+    def _check_sha(msg: dict, src: str, text: str, fail) -> None:
+        """Declared sha256 must match the file — guards against silent drift."""
+        declared = msg.get("sha256")
+        if not declared:
+            return
+        actual = hashlib.sha256(text.encode("utf-8")).hexdigest()
+        if actual != declared:
+            fail(f"sha256 drift for {src}: declared {declared[:12]}…, actual {actual[:12]}…")
```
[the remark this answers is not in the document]
adnanrhussain: Fixed in 8f5109c: prompt read is guarded (context manager) and sha256 now hashes raw bytes, so it's platform-independent.

evals/purpose_evaluator.ipynb
```diff
@@ -87,7 +87,7 @@
...
...predicted, expected)\n    results.append({\n        \"id\": fx[\"id\"], \"status\": status,\n        \"predicted\": predicted, \"expected\": expected,\n        \"description\": fx.get(\"description\", \"\"),\n    })\n\n# Per-case report\nprint(\"\\n\" + \"=\" * 78)\nprint(f\"{'ID':>5}  {'STATUS':<8}  {'PREDICTED':<22}  {'EXPECTED':<22}  DESCRIPTION\")\nprint(\"=\" * 78)\nfor r in results:\n    icon = {\"exact\": \"PASS\", \"adjacent\": \"PASS*\", \"fail\": \"FAIL\", \"error\": \"ERR\"}[r[\"status\"]]\n    print(f\"{r['id']:>5}  {icon:<8}  {(r['predicted'] or '-'):<22}  {r['expected']:<22}  {r.get('description','')[:25]}\")\n\n# Summary\nn_total = len(results)\nn_exact = sum(1 for r in results if r[\"status\"] == \"exact\")\nn_adj   = sum(1 for r in results if r[\"status\"] == \"adjacent\")\nn_fail  = sum(1 for r in results if r[\"status\"] == \"fail\")\nn_err   = sum(1 for r in results if r[\"status\"] == \"error\")\nprint(\"=\" * 78)\nprint(f\"Summary: {n_exact} exact, {n_adj} adjacent (tolerated), {n_fail} fail, {n_err} error  --  total {n_total}\")\nif _ALLOW_ADJ:\n    print(\"(Adjacency tolerance ON: predictions within +/-1 rubric step of the expected label count as PASS*.)\")"
```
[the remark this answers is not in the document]
adnanrhussain: Fixed in 8f5109c.

scripts/checks/eval_config.py
```diff
@@ -0,0 +1,164 @@
...
+
+    # --- Layer 2: cross-file -------------------------------------------------
+
+    def _check_referenced_files(self, config: dict, base: str, fail) -> None:
+        """The schema $refs for input/output must resolve to real files."""
+        for key in ("input_schema", "output_schema"):
+            ref = config.get(key, {})
+            if isinstance(ref, dict) and "$ref" in ref:
+                if not os.path.exists(os.path.join(base, ref["$ref"])):
+                    fail(f"{key} $ref not found: {ref['$ref']}")
```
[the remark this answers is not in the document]
adnanrhussain: Referenced input/output JSON validity is handled by the `eval-schemas` check in the stacked #127.

scripts/checks/eval_config.py
```diff
@@ -0,0 +1,164 @@
...
+        for ph in sorted(declared - template_vars):
+            fail(f'placeholder "{ph}" declared but not used in any prompt template')
+        for var in sorted(template_vars - declared):
+            fail(f'template variable "{{{var}}}" used but not declared in placeholders')
+
+    def _check_fixtures_path(self, config: dict, base: str, fail) -> None:
+        """Fixtures file must exist (its contents are validated by eval-fixtures)."""
+        path = config.get("fixtures", {}).get("path")
+        if path and not os.path.exists(os.path.join(base, path)):
+            fail(f"fixtures file not found: {path}")
```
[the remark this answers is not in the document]
adnanrhussain: Fixtures JSON validity is handled by the `eval-fixtures` check in the stacked #127.

evals/_schemas/config.schema.json
```diff
@@ -0,0 +1,170 @@
...
+  "description": "Shared contract for evals/**/config.json. One living schema, referenced by each config via its $schema field. Strict: additionalProperties is false at every level, so any new field must be added here deliberately.",
+  "type": "object",
+  "additionalProperties": false,
+  "required": ["$schema", "evaluator", "input_schema", "steps", "output_schema"],
+  "properties": {
+    "$schema": { "type": "string" },
+    "evaluator": {
+      "type": "object",
+      "additionalProperties": false,
+      "required": ["id", "name", "description", "supported_grades"],
```
czi-fsisenda, line 14: [P0] I don't think we should require supported_grades. We have seen evals that don't have grade as an input. input_schema is required and can enforce this.

evals/_schemas/config.schema.json
```diff
@@ -0,0 +1,170 @@
...
+      "properties": {
+        "id": { "type": "string", "minLength": 1 },
+        "name": { "type": "string", "minLength": 1 },
+        "description": { "type": "string", "minLength": 1 },
```
czi-fsisenda, line 18: [P1?] `"minLength": 1` feels like it implies a single character id, name, or description is ok. Can we go with `"required": true` or `"allow_empty": false` instead. Functionally the same, but no implicit endorsement of single character values for these fields.

evals/_schemas/config.schema.json
```diff
@@ -0,0 +1,170 @@
...
+        "supported_grades": {
+          "type": "array",
+          "minItems": 1,
+          "items": {
+            "enum": ["K", "1", "2", "3", "4", "5", "6", "7", "8", "9", "10", "11", "12"]
```
czi-fsisenda, line 23: [P1] I've been using int for grades with 0 = 'K'. I don't see a strong reason to prefer one over the other. Maybe ints are safer for mapping grade bands to grades? Either way, we should be consistent.
[P0] And from earlier comment, I think supported_grades should be an input field not an evaluator property.

evals/_schemas/config.schema.json
```diff
@@ -0,0 +1,170 @@
...
+    "steps": {
+      "type": "array",
+      "minItems": 1,
+      "items": {
+        "type": "object",
+        "additionalProperties": false,
+        "required": ["id", "type"],
+        "properties": {
+          "id": { "type": "string", "minLength": 1 },
+          "type": { "enum": ["llm", "api"] },
```
czi-fsisenda, line 91: [P1] Should we allow for a `"function"` type too? If there's a common step in a several evals that is not an llm or api call, we could still represent it as a step, e.g. Flesch-Kincaid.

evals/_schemas/config.schema.json
```diff
@@ -0,0 +1,170 @@
...
+                }
+              }
+            }
+          },
+          "model": {
+            "type": "object",
+            "additionalProperties": false,
+            "required": ["provider", "name"],
+            "properties": {
+              "provider": { "enum": ["google", "openai", "anthropic"] },
```
czi-fsisenda, line 132: [P1] Should we include `"custom`" even if we haven't implemented support yet?

evals/_schemas/config.schema.json
```diff
@@ -0,0 +1,170 @@
...
+            "required": ["messages", "placeholders"],
+            "properties": {
+              "messages": {
+                "type": "array",
+                "minItems": 1,
+                "items": {
+                  "type": "object",
+                  "additionalProperties": false,
+                  "required": ["role", "source_path"],
+                  "properties": {
```
czi-fsisenda, line 105: [P1] Newer models are dropping temperature, but most of the existing evals are using older models that have temperature as an attribute of the prompts step. Should we include `temperature` as part of the prompt properties? How are we capturing temperature at the moment?

evals/_schemas/config.schema.json
```diff
@@ -0,0 +1,170 @@
...
+              "provider": { "enum": ["google", "openai", "anthropic"] },
+              "name": { "type": "string", "minLength": 1 },
+              "alias": { "type": "string" }
+            }
+          },
+          "generation": {
+            "type": "object",
+            "additionalProperties": false,
+            "properties": {
+              "temperature": { "type": "number", "minimum": 0, "maximum": 1 }
```
czi-fsisenda, line 141: Ah! Here's where temperature lives.

evals/_schemas/config.schema.json
```diff
@@ -0,0 +1,170 @@
...
+                  "type": "object",
+                  "additionalProperties": false,
+                  "required": ["required", "source"],
+                  "properties": {
+                    "required": { "type": "boolean" },
+                    "source": { "type": "string" },
+                    "note": { "type": "string" }
+                  }
+                }
+              }
```
czi-fsisenda, line 124: [P0] Can't placeholders can be inferred from the prompt templates?
Langchain's approach lets users do partial placeholder replacements for templates but ensures all placeholders have been replaced before the template can be used in the prompt. If there are any un-replaced placeholders at prompt time, an exception is thrown.
I think forcing maintainers to define an attribute that can be inferred makes the framework tedious and creates opportunities for ambiguity, e.g. a placeholder is defined in the config, but is not present in the prompt template.

evals/_schemas/fixtures.schema.json
```diff
@@ -0,0 +1,19 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "fixtures.schema.json",
+  "title": "Evaluator fixtures",
+  "description": "Shared contract for evals/**/fixtures.json. Each case carries inputs and the expected label(s); per-evaluator input/expected shapes are validated against the evaluator's own input_schema/output_schema by the eval-fixtures check, not here.",
+  "type": "array",
+  "items": {
+    "type": "object",
+    "additionalProperties": false,
+    "required": ["id", "input", "expected"],
```
czi-fsisenda, line 10: [P1] The captures for the contract tests have: name (id), description, inputs, outputs, llm_captures. Should the fixtures replace the captures and include the additional fields?

evals/feedback/productive-coaching-writing-feedback/appropriate-feedback/config.json
```diff
@@ -1,9 +1,10 @@
...
+  "$schema": "../../../_schemas/config.schema.json",
   "evaluator": {
     "id": "feedback.productive_coaching_writing_feedback.is_appropriate_feedback",
     "name": "Appropriate Feedback Feedback-Quality Evaluator",
-    "description": "Did the feedback correctly identify whether the student needed to revise? (Key features generated from rubric text; concept paper did not list explicit key feature items.)"
+    "description": "Did the feedback correctly identify whether the student needed to revise? (Key features generated from rubric text; concept paper did not list explicit key feature items.)",
+    "supported_grades": ["8", "9"]
   },
   "input_schema": {
     "$ref": "input_schema.json"
```
czi-fsisenda, line 10: [P2] All the evals in a suite have the same input. Should we define that input once and reference it in all the evals?

evals/feedback/productive-coaching-writing-feedback/appropriate-feedback/config.json
```diff
@@ -1,9 +1,10 @@
 {
-  "spec_version": "1.0",
+  "$schema": "../../../_schemas/config.schema.json",
   "evaluator": {
     "id": "feedback.productive_coaching_writing_feedback.is_appropriate_feedback",
     "name": "Appropriate Feedback Feedback-Quality Evaluator",
-    "description": "Did the feedback correctly identify whether the student needed to revise? (Key features generated from rubric text; concept paper did not list explicit key feature items.)"
+    "description": "Did the feedback correctly identify whether the student needed to revise? (Key features generated from rubric text; concept paper did not list explicit key feature items.)",
+    "supported_grades": ["8", "9"]
```
czi-fsisenda, line 7: [P1] Grade isn't an input for the feedback evals. So it looks like `supported_grades` makes sense as an attribute of the eval. I am wondering how grade was captured for annotations though for us to determine that the eval is only appropriate for these grades.

evals/feedback/productive-coaching-writing-feedback/appropriate-feedback/example_notebook.ipynb
```diff
@@ -231,7 +231,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "fixtures_path = ASSETS_DIR / CONFIG[\"fixtures\"][\"sniff_test_path\"]\n",
+    "fixtures_path = ASSETS_DIR / CONFIG[\"fixtures\"][\"path\"]\n",
```
czi-fsisenda, line 234: [P0] Is this always the same? If it is, can we find it by convention instead of having the path as boilerplate?

evals/literacy/qualitative-text-complexity/intertextuality/fixtures.json
```diff
@@ -8,7 +8,7 @@
       "grade_level": 3
     },
     "expected": {
-      "complexity_level": "slightly_complex"
+      "complexity_score": "slightly_complex"
```
czi-fsisenda, line 11: [P2] Common outputs or output components for evaluator suites too?

evals/prompts/purpose/config.json
```diff
@@ -1,16 +1,18 @@
...
+    "description": "Evaluates the Purpose dimension of qualitative text complexity for K-12 reading assessment, producing a 5-level rubric rating with structured pedagogical detail.",
+    "supported_grades": ["3", "4", "5", "6", "7", "8", "9", "10", "11", "12"]
   },
   "input_schema": {
     "$ref": "input_schema.json"
   },
   "preprocessing": [
     {
       "id": "fk_score",
+      "type": "computation",
```
czi-fsisenda, line 15: [P1] Oh! We do have a "function" step. What is the benefit in having a distinction between `step` and `preprocessing`?

scripts/checks/eval_config.py
```diff
@@ -0,0 +1,179 @@
...
+
+Each config is validated in two complementary layers:
+
+  1. Schema  — structural rules a JSON Schema can express (required fields,
+     enums, types). Every config points at the shared contract via its own
+     `$schema` field; we load that and report every schema error.
+
+  2. Cross-file — rules a schema cannot express because they span files:
+       - referenced files actually exist ($ref schemas, prompt files, fixtures)
+       - each declared sha256 matches the prompt file on disk (drift tripwire)
```
czi-fsisenda, line 11: [P2] An alternative approach could be to generate a flat file from all the config jsons and have that represent a version of the eval. If the generated file changes, then we have a new version.
The generation step can be a proxy for a number of the checks. Generation fails if files are missing or are invalid.

scripts/checks/eval_config.py
```diff
@@ -0,0 +1,179 @@
...
+       - placeholders declared in config line up with the {vars} in the prompts
+       - system prompts carry no user-input placeholders
```
czi-fsisenda, line 13: [P0] Makes logical sense, but is it a necessary constraint? What if there is some valid reason to have a dynamic system prompt? This constraint may be too eager.

scripts/checks/eval_config.py
```diff
@@ -0,0 +1,179 @@
...
+       - no obsolete format-instruction placeholders survive anywhere
```
czi-fsisenda, line 14: [P2] Good for us to have automated enforcement for established patterns. But also wondering if this is too strict; having us lose flexibility. Could a warning work rather than a failure?

czi-fsisenda (approved):
Looks good! Approved not to block, but got a couple of comments.
Separately, the configs are clean and well structured, but a lot less ergonomic for tinkering with the notebooks.
It's not impossible to do experimentation with the notebooks. Just more complex.
We should check how users of the notebooks feel about the updates.

evals/_schemas/config.schema.json
adnanrhussain: You are right, but this isnt for the inputs. Its to describe which grades an eval is supported for. All our evals should be for K-12, but as we incrementally iterate and launch, the grade support matrix will fillin over cycles. For example, the new feedback evals are supported for grades 8-9 only. This field will help clarify that support

evals/_schemas/config.schema.json
adnanrhussain: `required` and `minLength` do different jobs: `required` enforces the key is *present*, `minLength: 1` enforces the *value* is non-empty. With `required` alone, `{"id": "", "name": ""}` validates fine - so `minLength: 1` is what actually rejects empty strings (it's the standard JSON Schema idiom for "non-empty," not an endorsement of 1-char values). `allow_empty` isn't a JSON Schema keyword. So we need both, and I'd keep them as-is.

evals/_schemas/config.schema.json
adnanrhussain: no strong opinion on either but I am leaning on string due to K and making it more readable

edit: see response on eval support vs input

evals/feedback/productive-coaching-writing-feedback/appropriate-feedback/config.json
adnanrhussain: @aychi1 - Would have the best insight on grade selection for feedback data

evals/prompts/purpose/config.json
adnanrhussain: nothing specific seperates them at the moment, if this shows up in the steps, we will merge these

evals/_schemas/config.schema.json
adnanrhussain: we do have a `computation` for for pre-processing. eventually we can support all across pre/steps/post, etc

evals/_schemas/config.schema.json
adnanrhussain: https://github.com/learning-commons-org/evaluators/pull/120/changes#r3502267749

evals/_schemas/config.schema.json
adnanrhussain: we can add in a followup

evals/_schemas/fixtures.schema.json
adnanrhussain: This is a generic placeholder, the inputs and outputs are the same as eval specific config.

Which properties should we also include here?

evals/feedback/productive-coaching-writing-feedback/appropriate-feedback/config.json
adnanrhussain: yeah, I was thinking about that, but didnt want to overoptimize until we have the full feedback suite implemented

evals/feedback/productive-coaching-writing-feedback/appropriate-feedback/example_notebook.ipynb
adnanrhussain: good point. i may have another improvement also, let me followup

scripts/checks/eval_config.py
adnanrhussain: Really like this, especially the versioning angle, let me come back to this

scripts/checks/eval_config.py
adnanrhussain: possibly, but its best practice for the system prompt to be static, and all user inputs are in the user prompt. Currently our prompts are prone to prompt injection due to this,  and need to be cleaned-up. Once enforced and the pattern reaches stability, we can relax this again if needed

scripts/checks/eval_config.py
adnanrhussain: the prompts shouldnt take an opinionated approach for format instruction placements in the prompts. Our users may not want it. This is not required, and more of a remnant that we need to cleanup

evals/_schemas/config.schema.json
czi-fsisenda: For input fields, `required` implies non-empty value, especially for strings. A form wouldn't allow `{"id": "", "name": ""}` if those were required fields. I think that is context we should be using here.

`required` on a form may allow `{"id": "  ", "name": "/n/n"}` though.
For the Python SDK, I went with `required` and `strip_whitespace` for attributes of text inputs.
https://github.com/learning-commons-org/evaluators/blob/a4bc33335d71192bc3c13256e50e7f7336ad3bea/sdks/python/src/learning_commons_evaluators/schemas/input_specs.py#L53-L80

evals/_schemas/config.schema.json
czi-fsisenda: I think I saw KG support pre-K as a grade. `-1` would be a confusing grade value.
`char` values may make more sense.

evals/_schemas/fixtures.schema.json
czi-fsisenda: We could add `description` and `captures` where `captures` have `id`, `type`, `request`, `response`.
Let's do this one later after doing some solid examples.

evals/_schemas/config.schema.json
czi-fsisenda: Saw your response below. Let's hear what Ariena says about how grade was determined for annotations.
I think I'm feeling more convinced that grade should be an input if the eval only works for some grade constraint. Counting on users checking an eval attribute seems like a hoping for best intentions where an explicit validation is possible.

PR #162, opened 2026-08-18
docs(sdks): language-agnostic SDK specification

adnanrhussain:
## Summary

Formalizes `sdks/SPEC.md` as the normative blueprint for all Evaluator SDKs — current (TypeScript, Python) and future. The spec is prescriptive: it defines the right contracts, and SDKs are brought into conformance with it.

### Structure

- **Design principles (§1)** — 11 principles every rule derives from (one canonical name; idiomatic surface / identical core; universal envelope; fail fast/loud/structured; observability never affects results; sensitive data opt-in; declared determinism; units in names; core vs. optional capabilities; executable spec)
- **Canonical contracts (§§2–8)** — naming & value forms, config, input validation, result envelope + payload rules, error taxonomy with data-driven retryability, logging, telemetry
- **Optional capabilities (§9)** — not required per SDK, but contract-bound when built (batch evaluation)
- **Evaluator registry (§10)** — evaluator definitions are shared, language-neutral data; definition schema, model pinning, prompt hygiene, derived inputs
- **Process & lifecycle (§§11–12)** — new-evaluator/new-SDK conformance checklist, contract fixtures as the executable spec, governance, spec versioning, compatibility & deprecation policy
- **Appendices** — A: known conformance gaps in current SDKs (snapshot 2026-08-18), B: per-SDK conformance matrix, C: open questions, D: changelog

### Key decisions

- One canonical name across languages, chosen to avoid builtins by construction (`RequestTimeoutError`, `InputValidationError`/`OutputValidationError` split)
- All durations in milliseconds with `_ms` suffix (`retry_after_ms`)
- Result envelope `{evaluator, result, metadata}` with per-evaluator payload shapes; profile taxonomy deliberately deferred (Q-2)
- Stability levels (Stable/Experimental) so the spec can move fast without breaking users
- `partner_key` → `partner_api_key`; new `telemetry.tracking_key` (semantics: Q-8)
- Hierarchical evaluator IDs (`literacy.ela_reading.vocabulary`), taxonomy: Q-7


sdks/SPEC.md
```diff
@@ -0,0 +1,499 @@
...
+
+## 3. Configuration
+
+Every evaluator accepts a config object at construction time.
+
+| Field | Type | Default | Description |
+|---|---|---|---|
+| `google_api_key` | string | none | Required by evaluators whose `default_providers` includes `google` |
+| `openai_api_key` | string | none | Required by evaluators whose `default_providers` includes `openai` |
+| `anthropic_api_key` | string | none | Required when `model_override.provider` is `anthropic` |
```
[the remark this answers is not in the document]
adnanrhussain: Fixed in 7b0557ba — all three provider key rows now use the symmetric form: required when `default_providers` includes the provider or `model_override.provider` targets it.

sdks/SPEC.md
```diff
@@ -0,0 +1,499 @@
...
+| `partner_api_key` | string | none | Learning Commons–issued partner key; forwarded as auth on telemetry requests, never in event bodies |
+| `model_override` | object | none | Override provider and model for all LLM calls (§3.2) |
+| `model_override.provider` | enum | — | A `Provider` value (§3.3) |
+| `model_override.model` | string | — | Pinned snapshot ID (§10.2) supported by that provider |
+| `max_retries` | int | `2` | Retry attempts on retryable errors (§6.3). Total attempts = 1 + `max_retries`; `0` disables |
+| `telemetry` | bool \| object | enabled | `true`/`false` shorthand, or an object for granular control |
```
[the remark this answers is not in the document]
adnanrhussain: Fixed in 7b0557ba — default is now `true`.

sdks/SPEC.md
```diff
@@ -0,0 +1,499 @@
...
+| `text_length_chars` | Trimmed input length |
+| `provider` | Model string(s) used (§3.4) |
+| `model_override` | `true` if an override was set; omitted otherwise |
+| `token_usage` | `{ input_tokens, output_tokens }` aggregated across phases |
+| `phase_details` | Array of per-phase objects (§8.3) |
+| `input_text` | Raw input — **only** when `telemetry.record_inputs = true` |
+
+### 8.3 Phase details
+
+Each entry: `phase` (name from the registry definition), `provider` (`"provider:model-id"`), `latency_ms`, `token_usage` (`{ input_tokens, output_tokens }`).
```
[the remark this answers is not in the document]
adnanrhussain: Fixed in 7b0557ba — renamed to `model` in both the event schema (§8.2) and `phase_details` (§8.3), consistent with `metadata.model` and the canonical-forms table. Added Appendix A gap 16 since both SDKs emit `provider` today.

sdks/SPEC.md
```diff
@@ -0,0 +1,499 @@
...
+1. Empty or whitespace-only → `InputValidationError("Text cannot be empty or contain only whitespace")`
+2. Trimmed length < min → `InputValidationError` stating the minimum
+3. Trimmed length > max → `InputValidationError` stating the maximum
+
+Messages MUST convey the same facts (which bound, what the bound is) and SHOULD match the canonical wording. `text_length_chars` in telemetry is the **trimmed** length.
+
+| Limit | Default |
+|---|---|
+| `min_text_length` | `10` |
+| `max_text_length` | `10,000` |
```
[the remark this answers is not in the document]
adnanrhussain: Fixed in 7b0557ba — `10000`.

sdks/SPEC.md
```diff
@@ -0,0 +1,500 @@
...
+
+New surfaces enter as Experimental and are promoted by maintainer agreement (§12.1) once at least one SDK ships them and contract fixtures (§11.3) cover them. Speed lives in Experimental; stability lives in Stable.
+
+---
+
+## 1. Design Principles
+
+Every rule in this spec derives from one of these. When the spec is silent, decide by principle, then codify the decision here.
+
+1. **One canonical name.** Every public identifier has exactly one canonical name, identical in every SDK and all documentation. Casing (§2.1) is the only permitted transformation; names are chosen to avoid known language builtins by construction (`RequestTimeoutError`, not `TimeoutError`).
```
czi-fsisenda, line 28: Had issues with this when implementing the Python SDK. Typescript used `ValidationError` which conflicted with Pydantics `ValidationError` to disambiguate them for Python we are using `InputValidationError`.
I think we can do best effort, but this may not always work.

sdks/SPEC.md
```diff
@@ -0,0 +1,500 @@
...
+
+2. **Idiomatic at the surface, identical at the core.** Types, async model, and packaging follow each language's conventions; behavior, names, and contracts are identical everywhere. When idiom and contract conflict, the contract wins and the gap is raised as a spec issue (§12.1).
+
+3. **Universal envelope, evaluator-scoped payload.** Every result shares one envelope; the domain payload is defined per evaluator in the registry (§10), under shared invariants (§5.2).
+
+4. **Fail fast, fail loud, fail structured.** Configuration problems surface at construction. Evaluation failures are canonical errors (§6) carrying retryability as data. Every error MUST be diagnosable from the error alone — silent fallbacks, swallowed causes, and generic messages are defects.
```
czi-fsisenda, line 34: 🙌

sdks/SPEC.md
```diff
@@ -0,0 +1,500 @@
...
+
+5. **Observability never affects results.** Telemetry and logging MUST NOT throw, block, delay, or change an evaluation's outcome.
+
+6. **No hidden state.** The full structured model output MUST be surfaced in every result — never empty, never omitted.
```
czi-fsisenda, line 38: No optional fields in output?

sdks/SPEC.md
```diff
@@ -0,0 +1,500 @@
...
+
+7. **Sensitive data is opt-in.** Raw user-supplied input text, and raw provider error strings (which may echo inputs or key fragments), MUST NOT appear in telemetry, logs, or error messages unless the caller explicitly opts in. Model outputs, scores, and reasoning are product data and MAY be logged and reported.
```
czi-fsisenda, line 40: I agree for telemetry and somewhat for logs, but I think we shouldn't filter error messages.
* Telemetry: We have a flag for including full inputs in telemetry so should it be MUST NOT unless explicitly and prominently opted-in for telemetry.
* Logs: We should be careful about logging in general, but can sensitive info show up in debug logs?
* Error messages: I think error messages should have as much detail as is available including sensitive info. The context for using the SDK is as code owner. The code owner should decide what to do with the error message. They can decide whether to log it, whether to include it in output, when and where it is available. If we sanitize the error, we limit the user's ability to diagnose issues. The CLI tool may need a different policy though.

sdks/SPEC.md
```diff
@@ -0,0 +1,500 @@
...
+## 3. Configuration
+
+Every evaluator accepts a config object at construction time.
+
+| Field | Type | Default | Description |
+|---|---|---|---|
+| `google_api_key` | string | none | Required when `default_providers` includes `google` or `model_override.provider` is `google` |
+| `openai_api_key` | string | none | Required when `default_providers` includes `openai` or `model_override.provider` is `openai` |
+| `anthropic_api_key` | string | none | Required when `default_providers` includes `anthropic` or `model_override.provider` is `anthropic` |
+| `partner_api_key` | string | none | Learning Commons–issued partner key; forwarded as auth on telemetry requests, never in event bodies |
```
czi-fsisenda, line 99: I think this should be `learning_commons_api_key` since that is what we are using for this.

sdks/SPEC.md
```diff
@@ -0,0 +1,500 @@
...
+| `model_override` | object | none | Override provider and model for all LLM calls (§3.2) |
+| `model_override.provider` | enum | — | A `Provider` value (§3.3) |
+| `model_override.model` | string | — | Pinned snapshot ID (§10.2) supported by that provider |
+| `max_retries` | int | `2` | Retry attempts on retryable errors (§6.3). Total attempts = 1 + `max_retries`; `0` disables |
+| `telemetry` | bool \| object | `true` | `true`/`false` shorthand, or an object for granular control |
+| `telemetry.enabled` | bool | `true` | Whether to emit telemetry events |
+| `telemetry.record_inputs` | bool | `false` | Include raw input text in telemetry (Principle 7) |
```
czi-fsisenda, line 106: `telemetry.include_full_inputs` or `telemetry.record_raw_inputs` or something like that?

sdks/SPEC.md
```diff
@@ -0,0 +1,500 @@
...
+| `telemetry.tracking_key` | string | none | Caller-supplied identifier attached to telemetry events, so a partner can segment usage by application or integration |
```
czi-fsisenda, line 107: `telemetry.telemetry_id` or `telemetry.telemetry_tracking_id`?

sdks/SPEC.md
```diff
@@ -0,0 +1,500 @@
...
+| `logger` | object | none | Custom logger implementing `Logger` (§7): `debug`/`info`/`warn`/`error`, each `(message, context?)`. Replaces the default console logger |
+| `log_level` | enum | `WARN` | Minimum level for the **default** logger: `DEBUG`, `INFO`, `WARN`, `ERROR`, `SILENT`. Ignored when a custom `logger` is provided |
```
czi-fsisenda, line 109: Python logging works differently. Loggers are namespaced in a hierarchy and, by convention, the SDK just emits to its named logger then the application owns handlers and levels.
Maybe `log_level` can be an implementation config rather than an SDK config?

sdks/SPEC.md
```diff
@@ -0,0 +1,500 @@
...
+| Class | Extends | Trigger | Retryable |
+|---|---|---|---|
+| `EvaluatorError` | — | Base class for all SDK errors | — |
+| `ConfigurationError` | `EvaluatorError` | Missing/invalid key, unknown provider or model, malformed settings | No |
+| `InputValidationError` | `EvaluatorError` | Caller-supplied text or grade failed validation (§4) | No |
+| `APIError` | `EvaluatorError` | Base class for failures in the provider call | per instance |
+| `AuthenticationError` | `APIError` | 401 / 403 from provider | No |
+| `RateLimitError` | `APIError` | 429 from provider | Yes |
+| `NetworkError` | `APIError` | Connection failure (DNS, refused, TLS) | Yes |
+| `RequestTimeoutError` | `APIError` | Provider request exceeded its timeout | Yes |
```
czi-fsisenda, line 220: I collapsed these into DependencyError in the latest version of the internal Python SDK. The reasoning is that this should be abstracted away from actual evaluation. Each integration with some external dependency should be wrapped where it is used so that the end-user only needs to understand if an issue is from inside the library or outside the library.
See https://github.com/chanzuckerberg/edu-evaluators/blob/c0747681da982e9dbce49bf5c799d405607f112b/sdks/python/docs/error-handling.md
Thoughts?

sdks/SPEC.md
```diff
@@ -0,0 +1,500 @@
...
+
+---
+
+## 3. Configuration
+
+Every evaluator accepts a config object at construction time.
+
+| Field | Type | Default | Description |
+|---|---|---|---|
+| `google_api_key` | string | none | Required when `default_providers` includes `google` or `model_override.provider` is `google` |
```
michael-berger-czi: seems like this list will get out of date fast if we don't remember to update this file when we add a new parameter. Should we remove, or move to the README and just reference the README here?

michael-berger-czi:
Is this spec for our reference as we develop our SDKs, or is this for our end users when they're evaluating the sdks?

adnanrhussain:
> Is this spec for our reference as we develop our SDKs, or is this for our end users when they're evaluating the sdks?

@michael-berger-czi - We are currently the primary builders and maintainers of the SDK. Eventually this model may evolve into a community-driven open-source approach, and so I would imagine codifying the spec is helpful for anyone contributing, or as you shared, trying to understand the SDK

michael-berger-czi (commented):
I suspect that a lot of things will become out of date (like parameters, event fields and output shape), and it will be an ongoing cost to remember to update this document as these low level details change. Recommend focusing this document on high level concerns, but open to discussing if you think otherwise.

sdks/SPEC.md
adnanrhussain, line 99: I wonder if we should rename this to `learning_commons_api_key` or `lc_api_key` to be explicit. What do folks think?

sdks/SPEC.md
```diff
@@ -0,0 +1,500 @@
...
+| `model_override.model` | string | — | Pinned snapshot ID (§10.2) supported by that provider |
+| `max_retries` | int | `2` | Retry attempts on retryable errors (§6.3). Total attempts = 1 + `max_retries`; `0` disables |
+| `telemetry` | bool \| object | `true` | `true`/`false` shorthand, or an object for granular control |
+| `telemetry.enabled` | bool | `true` | Whether to emit telemetry events |
+| `telemetry.record_inputs` | bool | `false` | Include raw input text in telemetry (Principle 7) |
+| `telemetry.tracking_key` | string | none | Caller-supplied identifier attached to telemetry events, so a partner can segment usage by application or integration |
+| `logger` | object | none | Custom logger implementing `Logger` (§7): `debug`/`info`/`warn`/`error`, each `(message, context?)`. Replaces the default console logger |
+| `log_level` | enum | `WARN` | Minimum level for the **default** logger: `DEBUG`, `INFO`, `WARN`, `ERROR`, `SILENT`. Ignored when a custom `logger` is provided |
+
+### 3.1 Key validation
```
czi-fsisenda, line 111: Can we add that key usage MUST require explicit intent. Config top level implies evaluation. Other uses need separate declaration. e.g. if we have `learning_commons_api_key` available in the env for KG calls, it doesn't mean that the user has opted into identified telemetry. If telemetry is going to use the key, we should have something separate like `telemetry.learning_commons_api_key = os.getEnv("LEARNING_COMMONS_API_KEY")`

sdks/SPEC.md
czi-fsisenda: Should we reserve this recommendation for terms we own directly, like eval names?

sdks/SPEC.md
```diff
@@ -0,0 +1,500 @@
+# Evaluator SDK Specification
+
+**Spec version:** 0.1.0 · **Changelog:** [Appendix D](#appendix-d-spec-changelog)
+
+This document is the normative blueprint for all Learning Commons Evaluator SDKs — current (TypeScript, Python) and future. It defines **what** every SDK must do, not how. Implementation details (retry libraries, async patterns, type systems) are left to each language's idioms.
+
+The spec is prescriptive, not descriptive: SDKs are brought into conformance with it, never the reverse. When the spec and an SDK diverge, the spec wins and the SDK carries the bug ([Appendix A](#appendix-a-known-conformance-gaps)).
+
+The key words **MUST**, **MUST NOT**, **SHOULD**, **SHOULD NOT**, and **MAY** are to be interpreted as described in RFC 2119.
```
czi-fsisenda, line 9: Should each of the items have one of these terms?

sdks/SPEC.md
```diff
@@ -0,0 +1,500 @@
...
+
+### Stability levels
+
+Every section and evaluator definition carries a stability level; unlabeled content is **Stable**.
```
czi-fsisenda, line 13: How do we label the features? Documentation (Mintlify?) or code tags or both?

sdks/SPEC.md
```diff
@@ -0,0 +1,500 @@
...
+
+### 3.1 Key validation
+
+- A key is *missing* if not provided or provided as the language's null/empty value.
+- If a required key (per the evaluator's `default_providers`, or the `model_override` provider) is missing, the constructor MUST raise `ConfigurationError` — before any I/O.
+- When `model_override` is set, **only** the override provider's key is required.
+
+### 3.2 `model_override`
+
+- The SDK MUST log a warning when an override is active — evaluators are validated against their default models only.
```
czi-fsisenda, line 119: Python SDK allows per instance overrides. Would we log the override once or for every evaluation? Maybe include override in evaluation info log line vs standalone warning?
For telemetry, makes sense to record override for every evaluation.

sdks/SPEC.md
```diff
@@ -0,0 +1,500 @@
...
+- Telemetry MUST record `model_override: true` (§8.2).
+- A model-not-found rejection from the provider raises `ConfigurationError`, not a retryable API error.
+
+### 3.3 `Provider`
+
+The closed set of supported LLM providers, used in `model_override.provider`, `default_providers`, and model strings (§3.4). Each SDK exposes it as its idiomatic closed enumeration.
```
czi-fsisenda, line 125: No support for custom / local models? Just for now?

sdks/SPEC.md
```diff
@@ -0,0 +1,500 @@
...
+
+1. Empty or whitespace-only → `InputValidationError("Text cannot be empty or contain only whitespace")`
+2. Trimmed length < min → `InputValidationError` stating the minimum
+3. Trimmed length > max → `InputValidationError` stating the maximum
+
+Messages MUST convey the same facts (which bound, what the bound is) and SHOULD match the canonical wording. `text_length_chars` in telemetry is the **trimmed** length.
+
+| Limit | Default |
+|---|---|
+| `min_text_length` | `10` |
```
czi-fsisenda, line 160: I think this should be 1. 10 is arbitrary.

sdks/SPEC.md
```diff
@@ -0,0 +1,500 @@
...
+| `max_text_length` | `10000` |
+
+An evaluator's registry definition (§10) MAY override these; overrides MUST NOT be silently ignored.
+
+### 4.2 Grade
+
+- Grade MUST be within the evaluator's declared supported range (§10) → otherwise `InputValidationError`.
+- Grade-free evaluators (e.g. GLA) MUST NOT silently consume a supplied grade.
```
czi-fsisenda, line 168: I think we should extend this to say that grade-free evaluators cannot have grade restrictions / recommendations. Annotation that didn't have grade as an input could not result in grade recommendations.
I think it's reasonable to say that a grade-free evaluator will only give results in a particular grade range though.

sdks/SPEC.md
```diff
@@ -0,0 +1,500 @@
...
+
+`RateLimitError` additionally: `retry_after_ms` (int | null) — milliseconds to wait, converted from the provider's `Retry-After` seconds.
+
+`OutputValidationError` additionally: `validation_errors` (array | null) — sanitized per-field failures (§6.4): schema locations and error types only.
+
+### 6.3 Retryability is data
+
+- `retryable` on the instance is the single source of truth; callers never memorize the hierarchy.
+- Resolution order: explicit per-instance override → `true` for any 5xx → class default.
+- The SDK's retry loop (`max_retries`, §3) retries **only** retryable errors, with exponential backoff, honoring `retry_after_ms` when present.
```
czi-fsisenda, line 243: OutputValidationErrors should retry immediately. Backing off in that case just increases eval latency.
I have a draft retry strategy recommendation implemented on the internal Python SDK that I'm working through. Almost fleshed out enough to share.

sdks/SPEC.md
```diff
@@ -0,0 +1,500 @@
...
+
+### 6.4 Error message safety
+
+- Messages MUST be short and controlled. Raw provider error strings MUST NOT be echoed into the message — they may contain prompt content, user text, or key fragments.
```
czi-fsisenda, line 247: This should only apply to SDK logging and telemetry. Any reason to limit info in exceptions messages?

sdks/SPEC.md
```diff
@@ -0,0 +1,500 @@
...
+| `tracking_key` | Caller-supplied `telemetry.tracking_key`; omitted when unset |
+| `grade` | Canonical grade token; null if the evaluator takes no grade |
+| `status` | `"success"` or `"error"` |
+| `error_code` | Canonical error class name on failure; null on success |
+| `latency_ms` | Total wall-clock time |
+| `text_length_chars` | Trimmed input length |
+| `model` | Model string(s) used (§3.4) |
+| `model_override` | `true` if an override was set; omitted otherwise |
+| `token_usage` | `{ input_tokens, output_tokens }` aggregated across phases |
+| `phase_details` | Array of per-phase objects (§8.3) |
```
czi-fsisenda, line 312: I think step is a clearer name. Evaluation phase vs evaluation step?

sdks/SPEC.md
```diff
@@ -0,0 +1,500 @@
...
+No SDK is required to build these — they exist where that SDK's users need them. Any SDK that ships one MUST implement the contract below (Principle 10). New capabilities are added here before the first SDK ships them.
+
+### 9.1 Batch evaluation *(Experimental)*
+
+- **Never throws for item failures.** Per-item failures are returned as structured entries alongside successes.
+- **Order-aligned results.** Results align index-for-index with inputs; failed items occupy their slot with the canonical error information (§6).
+- **Fail-fast only for batch-level faults.** Configuration errors and invalid batch arguments raise before any item is evaluated.
+- **Summary stats:** `total`, `successful`, `failed`, `average_processing_time_ms`.
+- **Per-item telemetry.** Each item fires its own event exactly as a single evaluation (§8); no batch-level rollup event.
+- **Bounded concurrency.** MUST bound concurrent provider calls; SHOULD expose the bound as an option.
```
czi-fsisenda, line 332: nit: Could we reword this or include an example?

sdks/SPEC.md
```diff
@@ -0,0 +1,500 @@
...
+
+Evaluator definitions are shared, language-neutral data: one definition per evaluator, consumed by every SDK, so evaluation behavior cannot drift between languages. Registry home and format: Q-6.
+
+### 10.1 Definition schema
+
+A definition MUST specify:
+
+| Field | Meaning |
+|---|---|
+| ID | Canonical evaluator ID — hierarchical, dot-namespaced, e.g. `literacy.ela_reading.vocabulary`. Appears in `evaluator` and `evaluator_type` |
```
czi-fsisenda, line 347: `id` mirrors folder structure?
Suggestion: `id` should only be namespaced by unique grouping, e.g. `qtc.vocabulary`. `literacy`, `ela`, `reading` should be metadata tags.

sdks/SPEC.md
```diff
@@ -0,0 +1,500 @@
...
+| Stability | Stable or Experimental |
```
czi-fsisenda, line 348: Should we have `deprecated` here too?

sdks/SPEC.md
```diff
@@ -0,0 +1,500 @@
...
+## 12. Spec Lifecycle
+
+### 12.1 Governance
+
+- The spec changes by pull request, approved by at least one maintainer of **each** SDK; small clarifications need one maintainer.
+- Substantive proposals SHOULD start as a short design note in the PR: motivation, alternatives, migration impact per SDK.
+- Every normative change adds a changelog line ([Appendix D](#appendix-d-spec-changelog)).
+- Open design questions live in [Appendix C](#appendix-c-open-questions) — promoted into the spec or closed with a recorded rationale, never silently dropped.
+
+### 12.2 Versioning
```
czi-fsisenda, line 414: Should we state that same eval spec should have same eval version in every SDK?
So we have SDK version that explicitly references language-agnostic SDK spec version and each eval has it's own version based on eval spec?
Do SDK versions have to be similar? e.g. if there's a new major version of a language and we need a new major version of SDK, is that fine?

sdks/SPEC.md
```diff
@@ -0,0 +1,500 @@
...
+
+- The spec is SemVer-versioned, independent of SDK packages. Pre-1.0, minor bumps may break; post-1.0, breaking changes to Stable content require a major bump.
```
czi-fsisenda, line 416: nit: Not clear which spec we are referring to.

sdks/SPEC.md
```diff
@@ -0,0 +1,500 @@
...
+
+Breaking is judged from the **user's** point of view:
+
+- **Non-breaking:** adding fields to results, events, or config; adding evaluators, error subclasses, or capabilities; widening accepted inputs; improving error messages within §4.1's same-facts rule.
+- **Breaking (major bump + deprecation):** removing or renaming any public name; changing a field's type, unit, or semantics; narrowing accepted inputs; changing a registry definition's models, phases, or temperature in a way that changes evaluation results.
+- Consumers of results, events, and registry data MUST treat unknown fields as forward-compatible (ignore, don't fail).
+
+**Deprecation:** a renamed or removed surface keeps a working, warning-emitting alias for at least one minor release of every affected SDK.
+
+**Evaluation-behavior changes** (model or prompt updates) are versioned in the registry and communicated in SDK changelogs even when the API is unchanged — users experience a score shift as a breaking change, whatever SemVer says.
```
czi-fsisenda, line 429: nit: Do we mean SDK when we say API here?

czi-fsisenda: So does eval SemVer drive SDK SemVer? If some eval has a breaking change, even if the SDKs have no other changes, then the SDKs all get a version bump?

sdks/SPEC.md
```diff
@@ -0,0 +1,500 @@
...
+| C-9 Divergences | ✓ | ✓ |
+| C-10 Spec version declared | ✗ | ✗ |
+
+---
+
+## Appendix C: Open Questions
+
+| # | Question | Notes |
+|---|---|---|
+| Q-1 | Should SDKs read API keys from environment variables as a fallback to explicit config? | UX win for scripts vs. surprise credential pickup in servers; precedence and naming must be canonical if adopted |
```
czi-fsisenda, line 485: Suggestion: Always read keys from explicit config AND config construction can fallback to env variables. That way users can trust config. OR remove all keys from config and always read keys from env variables.

sdks/SPEC.md
```diff
@@ -0,0 +1,500 @@
...
+| Q-2 | Payload shape taxonomy — should evaluator families (QTC, feedback, standards alignment, …) share declared payload profiles? | Deliberately deferred (§5.2); revisit once feedback and math evaluator classes mature |
+| Q-3 | Should registry definitions carry a version (prompt/model revision) surfaced in result `metadata`? | Would let users pin/detect evaluation-behavior changes (§12.3) programmatically |
```
czi-fsisenda, line 487: +1

czi-fsisenda (approved):
Nice! 🚀
Lots of comments, but just because we are covering so much here.
I'm sure this will evolve a lot as use cases we never imagined come up.
Only 2 issues I feel strongly about for this iteration:
1. Full exception stack trace must be available to sdk users. Exception error messages should be as descriptive as possible - even including sensitive info if necessary. Agree that SDK doesn't log sensitive info or include sensitive info in telemetry.
2. OutputValidationErrors should retry immediately. No backoff.

sdks/SPEC.md
czi-fsisenda: Worded differently, grade guidance for an evaluator only makes sense if the evaluator has grade as an input or output. Grade guidance range has to match the input / output grade range.

sdks/SPEC.md
adnanrhussain: Clarified

sdks/SPEC.md
adnanrhussain: agreed. updated

sdks/SPEC.md
adnanrhussain: used record_raw_inputs, and is preferable since we may have more than 1 text inputs (feedback evals) and non-textual inputs (math visual eval)

sdks/SPEC.md
adnanrhussain: going with `telemetry.learning_commons_api_key` instead, to follow convention and be very explicit.

sdks/SPEC.md
adnanrhussain: yes. updated those two into separate explicit keys

sdks/SPEC.md
adnanrhussain: log once per initialization, not invocation. Telemetry includes override flag

sdks/SPEC.md
adnanrhussain: not yet. we will come back to this

michael-berger-czi (approved):

sdks/SPEC.md
```diff
@@ -0,0 +1,535 @@
...
+### 6.4 Trust boundaries
+
+The SDK caller is a code owner who already holds the API keys and inputs; diagnostic detail is theirs. What is limited is what leaves the process by default.
+
+| Surface | Policy |
+|---|---|
+| Exception attributes + cause chain | Full diagnostic detail MUST be preserved: original exception and stack trace via the language's cause chain, plus structured attributes (§6.2) and allowlisted upstream diagnostics (provider message/code/reason) when structured payloads carry them |
+| Error message | MUST be diagnosable and actionable on its own; MAY include input-derived detail where it aids diagnosis; MUST NOT contain credentials or key fragments — messages propagate by default (app logs, error trackers, HTTP responses) |
+| Logs | No raw input text above `DEBUG` (§7) |
+| Telemetry | Error class name and sanitized schema facts (field paths, error types) only; raw input only via `telemetry.record_raw_inputs` (§8) |
```
michael-berger-czi, line 283: @adnanrhussain. Guessing we should remove record_raw_inputs? Also should this be merged into main? I might be able to iterate on it while I do the python sdk changes.

PR #164, opened 2026-08-19
chore(evals-prompts): remove text_complexity_combo.ipynb

adnanrhussain:
Removes `evals/text_complexity_combo.ipynb` — high maintenance cost, low value.

## Why

- Its vocabulary integration is already broken on `main`: it calls `v_prompts.SYSTEM_PROMPT`/`USER_PROMPT`, attributes `vocab_prompts.py` hasn't exposed since it moved to a grade-keyed dict. Would `AttributeError` if run today.
- Duplicates prompt-loading logic each evaluator's own dedicated notebook already has, drifting independently.
- "Early Release" side artifact; nothing else depends on it.

## Scope

Deletes the notebook (1152 lines) and its row in `evals/README.md`. `sent_str_prompts.py`/`smk_prompts.py` keep their own dedicated-notebook consumers — nothing orphaned.

## Merge order

Recommend merging this **before** #161/#163 — both currently patch this notebook's import cell differently, which is a real merge conflict between them. If this merges first, each just hits a simple modify/delete conflict against `main` instead (keep the deletion).

## Verification

`scripts/check.py` passes. No SDK impact.


michael-berger-czi (approved):

adnanrhussain:
@esthersweon - FYI. I am deleting this python notebook, and I checked documentation to confirm there are no references / broken links.

PR #169, opened 2026-08-20
fix(evals): allow id_history to be shared across evaluators for legitimate splits

adnanrhussain:
## Summary
Fixes #168's uniqueness check to allow `id_history` values to repeat across evaluators, for the legitimate case of one evaluator splitting into several. Vocabulary is becoming two evaluators (grades-3-4, other-grades) from one historical shipped id (`vocabulary`), and both need that id in their `id_history` for accurate bidirectional lookup -- the original check would have flagged that as reuse.

## New behavior
- `stable_id` and current `id`: still strictly unique across all evaluators (unchanged).
- `id_history` entries: may repeat across evaluators (a shared retired id represents shared ancestry from a split).
- Still caught: a retired id (in `id_history`) becoming a *different* evaluator's live, current `id` -- that's a real collision, not a split.

## Verification
`python3 scripts/check.py` passes. Manually verified: a shared `id_history` value across two evaluators now passes; a retired id colliding with another evaluator's current id still fails; duplicate current `id` still fails; the #168 crash-guard (non-dict `evaluator`) still holds.

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -36,7 +36,7 @@ fails the PR (it never auto-fixes) — so fixing locally saves a round-trip.
 | Check | What it does |
 |-------|--------------|
 | `strip-notebooks` | Clears outputs, execution counts, and cell metadata from `.ipynb` files |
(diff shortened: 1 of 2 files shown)


adnanrhussain:
Superseding this — we're merging Vocabulary's two configs (grades-3-4, other-grades) into one evaluator instead of keeping the split, so there won't be two evaluators sharing a historical id. The strict 1:1 uniqueness check from #168 stays as-is.

PR #163, opened 2026-08-18
feat(evals-prompts): add Vocabulary Complexity onto the shared evaluator contract

adnanrhussain:
Adds the Vocabulary Complexity evaluator (formerly named Vocabulary) at `evals/student-facing-text/ela-reading/vocabulary-complexity/`, following the same shared contract as Grade Level Appropriateness (#159) and Meaning Directness (#161).

## Evaluator

- `evaluator.id`: `student_facing_text.ela_reading.vocabulary_complexity`
- `stable_id`: `d6f8e6d1-3d0d-4b6c-9a3a-2b7c9e6f1a4d`, `id_history`: `["vocabulary"]` (its real shipped SDK id)
- One `config.json`, three LLM steps:
  - `background_knowledge` — always runs, shared by both grade bands (GPT-4o)
  - `vocab_complexity_grades_3_4` — `condition: grade_level in [3,4]` (Gemini 2.5 Pro), with an `fk_score` preprocessing step gated by the same condition
  - `vocab_complexity_other_grades` — `condition: grade_level in [5..12]` (GPT-4.1)
- Internal grade-band-prefixed filenames (`grades-3-4-*`, `other-grades-*`) reflect grade band, not evaluator name, and are unchanged
- `grade_level` is a string across `input_schema.json`, `config.json` placeholders, and fixtures, matching `evaluator.supported_grades`'s own convention

## Contract

- `output_schema.json` shared by both grade-band branches: `tier_2_words`, `tier_3_words`, `archaic_words`, `other_complex_words`, `complexity_score`, `reasoning`
- `fixtures.json` + `example_notebook.ipynb` evaluate each step's `condition` generically against the input to select the right branch, then run the config end-to-end against real LLMs

## Scope

TS SDK class/id/field names are unchanged in this PR — only the eval's prompt import paths were updated to point at the new folder. SDK-side renames are deferred to a follow-up pass.

## Verification

`scripts/check.py` all pass · TS `tsc --noEmit` clean.

PR #172, opened 2026-08-20
refactor(evals): merge Vocabulary's two grade-band configs into one evaluator

adnanrhussain:
WIP -- verifying live CI before finalizing. Depends on #171 (condition field).


adnanrhussain:
Wrong location -- this work belongs directly on #163, which it supersedes. Moved there instead of creating a parallel PR.

PR #177, opened 2026-08-20
feat(evals): add Reference Knowledge Demands onto the shared evaluator contract

adnanrhussain:
Adds the Reference Knowledge Demands evaluator (formerly named Intertextuality) at `evals/student-facing-text/ela-reading/reference-knowledge-demands/`, following the same shared contract as Grade Level Appropriateness (#159), Meaning Directness (#161), Vocabulary Complexity (#163), Background Knowledge Demands (#173), Sentence Structure (#174), Purpose Clarity (#175), and Organizational Structure (#176). This evaluator already had a full contract (`config.json` + schemas + 4 real fixtures + `example_notebook.ipynb`) at `evals/literacy/qualitative-text-complexity/intertextuality/`, just not yet at the new taxonomy path or under the new name.

## Evaluator

- `evaluator.id`: `student_facing_text.ela_reading.reference_knowledge_demands`
- `stable_id`: `b52b1924-73b7-40e3-bbca-081302c83f30`, `id_history`: `["literacy.gla.intertextuality"]` (its previous id)
- Single LLM step (`evaluate_intertextuality`, kept as-is — an internal step id, not evaluator-facing) on Gemini 3 Flash Preview, plus an `fk_score` preprocessing step
- `grade_level` is now a string enum (`["3"..."12"]`) across `input_schema.json`, `config.json` placeholders, and fixtures — was previously an integer with `minimum`/`maximum`
- `evaluator.name` renamed to "Reference Knowledge Demands Evaluator", matching the sibling naming convention

No published SDK (TS or Python) consumes this evaluator's `config.json` today, so there's no downstream coupling to worry about — unlike Purpose Clarity (#175).

## Contract

- `output_schema.json` defines the structured output: `complexity_score` (4-level rubric), `reasoning`, `details` (`detailed_summary`, `adjustment_and_scaffolding`, `recommended_use_cases`); the `$defs.IntertextualityDetails` schema name renamed to `ReferenceKnowledgeDemandsDetails` for consistency (pure internal schema naming — no external consumer)
- `fixtures.json` retains its 4 existing real fixture cases, with `grade_level` updated to match the new string type
- `example_notebook.ipynb` already matched this repo's standard config-driven, hash-verified pattern — updated the title/identity markdown, `grade_level` type hint/sample call to string, and fixed a stale path reference in a docstring. The prompt content in `system.txt`/`user.txt` is untouched.

## evals/README.md

Added a new row for this evaluator — it didn't have one before.

## Verification

`scripts/check.py` all pass.

PR #173, opened 2026-08-20
feat(evals): add Background Knowledge Demands onto the shared evaluator contract

adnanrhussain:
Adds the Background Knowledge Demands evaluator (formerly named Subject Matter Knowledge / SMK) at `evals/student-facing-text/ela-reading/background-knowledge-demands/`, following the same shared contract as Grade Level Appropriateness (#159), Meaning Directness (#161), and Vocabulary Complexity (#163). Previously this evaluator only had loose prompt files under `evals/prompts/subject-matter-knowledge/` with no config, schemas, or fixtures.

## Evaluator

- `evaluator.id`: `student_facing_text.ela_reading.background_knowledge_demands`
- `stable_id`: `60ddf4c3-8425-40cc-bb42-a3975e20fd83`, `id_history`: `["subject-matter-knowledge"]` (its real shipped TS SDK id)
- Single LLM step (`evaluate_background_knowledge_demands`) on Gemini 3 Flash Preview, plus an `fk_score` preprocessing step (Flesch-Kincaid grade level)
- `grade_level` is a string across `input_schema.json`, `config.json` placeholders, and fixtures, matching `evaluator.supported_grades`'s own convention. `system.txt`'s own input-data description previously said `(integer)`; fixed to `(string)` to match. No other prompt content changed — the rubric itself (still internally called "Subject Matter Knowledge (SMK)" in the prompt text) is untouched.

## Contract

- `system.txt` / `user.txt` prompt files, hash-pinned in `config.json`
- `output_schema.json` defines the structured output: `complexity_score` (4-level rubric), `identified_topics`, `curriculum_check`, `assumptions_and_scaffolding`, `friction_analysis`, `reasoning`
- `fixtures.json` + `example_notebook.ipynb` run the config end-to-end against a real LLM, with adjacent-level fixture tolerance. The one fixture case is drawn directly from a calibration example embedded in the rubric's own prompt text.

## Scope

TS SDK class/id/field names (`SmkEvaluator`, `subject-matter-knowledge` id, `SmkOutputSchema`) are unchanged in this PR — only the eval's prompt import paths were updated to point at the new folder, and the user-prompt substitution key changed from `{grade}` to `{grade_level}` to match the new canonical prompt. SDK-side renames are deferred to a follow-up pass.

## Verification

`scripts/check.py` all pass · TS `tsc --noEmit` clean.

PR #159, opened 2026-08-17
fix(evals-prompts): correct GLA grade bands and re-sync prompt copies

adnanrhussain:
Migrates Grade Level Appropriateness onto the shared evaluator contract (`config.json` + schemas + fixtures), fixes a real cross-SDK bug, and moves it to the new `evals/<domain>/<skill-modality>/<evaluator>/` folder taxonomy.

## Breaking changes

- Output fields renamed: `grade` → `grade_band`, `alternative_grade` → `alternative_grade_band` (both are bands like `"4-5"`, never single grades).
- Path: `evals/literacy/ela-reading/grade-level-appropriateness/`. `evaluator.id`: `literacy.ela_reading.grade_level_appropriateness`.

## What's fixed

- **The TS and Python SDKs disagreed on the top grade band** — TS shipped `11-12` (a typo in `user.txt`), Python's enum already said `11-CCR` (matching `system.txt`). Fixed to `11-CCR`; added the missing `K-1` word-count row.
- All prompt instructions now live in `system.txt`; `user.txt` just takes `{text}`. `{format_instructions}` removed — structured output (driven by `output_schema.json`) replaces it.
- `gla_prompts.py` deleted (its only consumer was the old notebook, now rewritten to load `config.json` + prompts from disk).
- Model bumped to `gemini-3.6-flash` @ temperature 0.

## Follow-ups (not in this PR)

- Both SDKs need the `grade_band` rename applied to their own schemas — until then TS's prompt and its Zod schema disagree on the field name (structured output means it won't crash, but it's wrong).
- Python's vendored prompt copy (`sdks/settings/...`) needs `make sync-settings`.
- Fixture coverage is one case (from an SDK contract capture); real coverage is a follow-up.

## Verification

`scripts/check.py` all pass · TS `tsc --noEmit` clean · 322 unit tests pass · `npm run build` succeeds and the bundle contains the fix. Notebook itself not executed (needs live API keys).

## Note: grade_level standardization (Conventionality, Vocabulary)

The repo settled on `grade_level` as the standard placeholder/field name for a single grade (previously `grade`, `grade_level`, `student_grade_level` were all in use across different evaluators — see #161, #163). Not applicable here: GLA takes no grade input at all (`text` only) and outputs `grade_band`/`alternative_grade_band`, a different concept.

## Update: capture mode + notebook execution CI (#165)

[...]

PR #185, opened 2026-08-21
feat(evals): add Strength Acknowledgment onto the ela-writing taxonomy

adnanrhussain:
Migrates Acknowledges Strength onto the `evals/feedback/ela-writing/` taxonomy as **Strength Acknowledgement**, at `evals/feedback/ela-writing/strength-acknowledgement/`. Same pattern as the student-facing-text family (#159, #161, #163, #173–#177). The evaluator already had a complete contract; this moves it, renames it, and closes the gaps that family standardised on.

## Evaluator

- `evaluator.id`: `feedback.ela_writing.strength_acknowledgement`
- `stable_id`: `99370b62-6791-4535-a77e-dda6806c635c`, `id_history`: `["feedback.productive_coaching_writing_feedback.is_acknowledges_strength"]`
- `evaluator.name`: `Strength Acknowledgement Evaluator` — drops the `Feedback-Quality` suffix so it matches the `<X> Evaluator` form used across student-facing-text
- Step id `evaluate_is_acknowledges_strength` → `evaluate_strength_acknowledgement`
- Single LLM step on `gpt-5.4-2026-03-05`, temperature 1, unchanged

## `quality_score`

The output score field was named after the evaluator (`acknowledges_strength_score`). Every evaluator in this family had a differently-named field for the same binary judgement, so a consumer had to know which evaluator ran before it could read the result.

Student-facing-text names this after the measured *dimension* — `complexity_score`, identical across all 8 of its evaluators. This adopts the same convention: **`quality_score`**, shared across all 7 feedback evaluators. Values are unchanged (`integer`, `0`/`1`).

Renamed in `output_schema.json` (property, `required[]`, description), `fixtures.json` (`expected`), and the notebook. Nothing consumes these configs — no TypeScript or Python SDK reads them — so there is no downstream breakage.

## Also fixed

The notebook carried two copy-paste references to `evals/prompts/purpose`, a leftover from the Purpose evaluator it was templated from. Both now point at this directory.

## Not applicable

The `grade_level` string standardisation doesn't apply — this evaluator takes `student_text` and `feedback_text`, no grade input. `supported_grades: ["8", "9"]` is metadata only.

## Verification

`scripts/check.py` all pass.

## Spelling: "acknowledgement"

The public name uses `acknowledgement`; the repo used `acknowledgment` in 14 places. Both are valid — the former is the British variant, the latter standard US — so this is a variant choice, not a typo fix. Standardised on `acknowledgement` throughout, **including `system.txt`**.

Editing prompt text is deliberately out of scope for a rename elsewhere in this migration series, so calling it out: it was authorised here specifically, because leaving the prompt on the other spelling would make the evaluator internally inconsistent with its own name and folder. The pinned `sha256` for `system.txt` is recomputed to match (`d66c26d4…` → `3a96d931…`), and `scripts/check.py` verifies it.


aychi1 (approved):
LGTM, conditional on P0 max length.

PR #188, opened 2026-08-22
feat(evals): add Critical Thinking onto the shared evaluator contract

adnanrhussain:
Adds the Critical Thinking evaluator at `evals/durable-skills/ela-writing/critical-thinking/`, following the same shared contract as the student-facing-text and feedback families (#159, #161, #163, #173–#177, #179–#185). First evaluator under the `durable-skills` domain.

Contract-only: the TypeScript evaluator follows in its own PR, so the rubric, schemas and fixtures can be reviewed by the people who own them first.

**Scope: grades 8-10 only.** `supported_grades` is `["8", "9", "10"]`, and the rubric prompt states the same range in both places it appears. The validation evidence covers those grades and no others, so the evaluator is not offered for grades 6-7 or 11-12.

## Evaluator

- `evaluator.id`: `durable_skills.ela_writing.critical_thinking`
- `stable_id`: `114fb24f-65d8-4736-bb23-300b0470fc4d`
- Single LLM step (`evaluate_critical_thinking`) on `claude-opus-5`, `provider: anthropic`
- Inputs are `assignment_text`, `sources`, `source_count`, `essay_text` — no grade input, so the `grade_level` string standardisation doesn't apply
- `supported_grades` is `8`–`10` (see Scope above); the prompt's two grade references match
- Rates five indicators (synthesizing sources, evidence strength, counterarguments, facts over opinions, drawing conclusions) plus an Overall, on a five-level scale — `not_evident`, `exploring`, `analyzing`, `integrating`, `extending`. `synthesizing_sources` is omitted when `source_count` is 1

## Contract

- `system.txt` / `user.txt`, hash-pinned in `config.json`
- `output_schema.json` defines `indicators` (each with `evidence`, `rationale`, `rating`) then `overall`. Property order is deliberate — evidence and rationale precede each rating, and all indicators precede the Overall, so the reasoning conditions the score
- **`indicators` is an object keyed by indicator id**, not an array. `evidence_strength`, `counterarguments`, `facts_over_opinions` and `drawing_conclusions` are `required`; `synthesizing_sources` is an optional fifth, present only on multi-source prompts. `additionalProperties: false` closes the rest. Nothing in the prompts required an array, and named keys say all of this in the schema itself rather than in `allOf`/`contains` — and make duplicate ids impossible, which no array constraint could
- `Indicator.id` is gone: the key carries the identity, so keeping both would let them disagree. The id-to-rubric mapping (2.1, 2.2, 3.1, 3.2, 4.1) moved to the `indicators` description
- `fixtures.json` carries two real cases covering both branches: one multi-source with all 5 indicators, one single-source with 4. Each case's `description` records the per-indicator expert consensus it was labelled from
- `example_notebook.ipynb` runs the config end-to-end, verifying both prompt hashes on load and deriving its generation settings from the config

## `generation.temperature` is `null`

First use of the nullable temperature from #186. Verified against the live API on `claude-opus-5`:

| sent | result |
|---|---|
| omitted | 200 |
| `1` | 200 |
| `0` | **400 — "`temperature` is deprecated for this model"** |
| `0.5` | **400** |
| literal `null` | **400 — "Input should be a valid number"** |

Anthropic's guidance for Opus 4.7 and later is to omit the parameter entirely, which is exactly what `null` means in our contract.

`null` is also the only correct encoding here, not merely a convenient one: `config.schema.json` requires `generation.temperature`, and in the SDK an *absent* value falls back to the evaluator config's default while `null` explicitly omits. Removing the key would send a temperature rather than suppress one.

The model is `claude-opus-5` with no dated suffix. Per Anthropic's versioning docs, dateless ids from the 4.6 generation onward are themselves pinned snapshots rather than evergreen pointers, so there is no dated form to pin instead.

## Also fixed

- **`$schema`** was one level short for a three-deep path — `eval-config` caught it
- **`langchain-anthropic` added to `evals/requirements.txt`** — this is the first eval to use it, and the existing `anthropic` entry is the raw SDK, not the LangChain integration. `eval-requirements` caught it
- **`fixtures.tolerance` added** — the notebook already read `tolerance.allow_adjacent_levels`, but config never set it, so fixtures ran on a strict gate. Now matches the sibling evaluators
- Notebook switched to raw-bytes prompt hashing, matching the rest of the repo
- **The notebook now derives `temperature` from the config** rather than printing it and ignoring it. It previously built the model without a temperature, which was correct for this config by coincidence — `null` means omit and it omitted — but any config that set a number would have been silently disregarded
- **`registry-conformance.test.ts`** gains a `UNIMPLEMENTED` entry for this id. That test asserts every contract on disk is either implemented or explicitly named, so a contract-only PR has to declare itself; the entry says to drop it when the TypeScript evaluator lands. This is the only file touched outside the eval directory

## Verification

`scripts/check.py` — all 6 pass. TypeScript unit tests, typecheck and lint pass, which this PR sees for the first time: it predates #223, so the old path filter (`evals/prompts/**`, `evals/student-facing-text/**`) never matched a new domain and the SDK jobs never ran.

The notebook was executed against the real API using the same command and timeout as `eval-notebooks.yml`. Exit 0, no cell errors, both fixtures exact:

| fixture | predicted | expected | status |
|---|---|---|---|
| `5076429_cleaned` (multi-source, 5 indicators) | `analyzing` | `analyzing` | exact |
| `AAAOPP13416000111182` (single-source, 4 indicators) | `exploring` | `exploring` | exact |

Both matched exactly rather than relying on the adjacency tolerance. Re-run after each change: after the rating values moved to snake_case, after the level-token hint was removed from the prompt (confirming the schema enum alone constrains the model), and after `indicators` became an object. The live payload came back keyed by id with all five indicators and no `id` field.

A live run exercises one branch at a time, so the schema's accept/reject behaviour is checked directly as well:

| case | result |
|---|---|
| single-source: four mandatory only | accepted |
| multi-source: all five | accepted |
| missing a mandatory indicator | rejected |
| unknown indicator id | rejected |
| old Title Case rating | rejected |
| redundant `id` field | rejected |
| `evidence: []` at `analyzing` | **accepted — the known gap below** |

## Open items, not blocking

- **`evidence` may be empty at any rating.** The prompt requires at least one quote for anything above `not_evident`, but the schema permits an empty array, so the structured-output parser will not enforce it. Expressible as an `if`/`then` on `rating`; not done here.
- **`outcome` is not declared**, so `readOutcome` will return no score for this evaluator once it is implemented. Every scored evaluator except math declares one; worth adding when the SDK side lands.
- **`fixtures.tolerance.allow_adjacent_levels` is read only by the notebook.** Nothing in the SDK or the checks consumes it.
- **`max_tokens=16000` is hardcoded in the notebook.** `generation` only permits `temperature` (`additionalProperties: false`), so there is nowhere in the contract to declare it. Measured output is ~4.5K tokens, so the ceiling is generous headroom rather than a tuned value.
- **Expert sign-off on the output shape is still open** — see the thread on `output_schema.json`.
- No `evals/README.md` row, consistent with the other in-flight evaluator PRs.


evals/durable-skills/ela-writing/critical-thinking/fixtures.json
```diff
@@ -0,0 +1,36 @@
+[
+  {
+    "id": "5076429_cleaned",
+    "description": "Multi-source prompt (2.1 rated). Clean hit: evaluator matches expert consensus on all indicators. Expert consensus (Round 3, 6 raters): 2.1=Analyzing, 2.2=Analyzing, 3.1=Analyzing, 3.2=Analyzing, 4.1=Analyzing, Overall=Analyzing.",
+    "source": "PERSUADE 2.0 (CLEAR corpus), CC BY 4.0. Label = Round 3 expert consensus (round-up median of rater ratings).",
```
adnanrhussain: May need an update to our license information

evals/durable-skills/ela-writing/critical-thinking/output_schema.json
```diff
@@ -0,0 +1,111 @@
...
+    "overall": {
+      "$ref": "#/$defs/Overall",
+      "description": "The headline Critical Thinking rating, formed last as the median of the indicator ratings."
+    }
+  },
+  "$defs": {
+    "Level": {
+      "type": "string",
+      "enum": [
+        "Not Evident",
```
adnanrhussain: This may need to be updated to snake_case for consistency. May require updates to the model prompt

evals/durable-skills/ela-writing/critical-thinking/output_schema.json
```diff
@@ -0,0 +1,111 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "output_schema.json",
+  "title": "EvaluatorOutput",
+  "description": "Critical Thinking ratings for a student argumentative essay. Property order is intentional: each indicator's evidence and rationale are produced before its rating, and all indicators are produced before the Overall rating, so reasoning conditions the ratings.",
+  "type": "object",
```
adnanrhussain, line 6: @aychi1 - Did we align with our internal experts on the output shape of this Eval?

evals/durable-skills/ela-writing/critical-thinking/config.json
```diff
@@ -0,0 +1,81 @@
+{
+  "$schema": "../../../_schemas/config.schema.json",
+  "evaluator": {
+    "id": "durable_skills.ela_writing.critical_thinking",
+    "stable_id": "114fb24f-65d8-4736-bb23-300b0470fc4d",
+    "name": "Critical Thinking Evaluator",
+    "description": "Rates grade 6-12 student argumentative essays for Critical Thinking across five indicators (synthesizing sources, evidence strength, counterarguments, facts over opinions, drawing conclusions) plus an Overall rating, on a five-level scale (Not Evident to Extending).",
```
aychi1: "Grades 6-12" --> "Grades 8-10"

evals/durable-skills/ela-writing/critical-thinking/config.json
```diff
@@ -0,0 +1,81 @@
...
+    "supported_grades": [
+      "6",
+      "7",
+      "8",
+      "9",
+      "10",
+      "11",
+      "12"
```
aychi1: "supported_grades": [8,9,10]

evals/durable-skills/ela-writing/critical-thinking/fixtures.json
```diff
@@ -0,0 +1,36 @@
...
+    "input": {
+      "source_count": 3,
```
aychi1, line 6: As a sanity check-- what is the source for this?

evals/durable-skills/ela-writing/critical-thinking/output_schema.json
aychi1: Hi @adnanrhussain, I personally did not.

evals/durable-skills/ela-writing/critical-thinking/output_schema.json
```diff
@@ -0,0 +1,111 @@
...
+  "additionalProperties": false,
+  "properties": {
+    "indicators": {
+      "type": "array",
+      "minItems": 1,
+      "description": "One entry per rated indicator. Indicator 2.1 (synthesizing_sources) is rated only when source_count > 1; when source_count is 1 it is omitted from this array entirely.",
+      "items": {
+        "$ref": "#/$defs/Indicator"
+      }
+    },
```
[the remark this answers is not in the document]
aychi1: Agree with this feedback.

aychi1 (commented):
Overall LGTM. Several minor questions/ comments.

evals/durable-skills/ela-writing/critical-thinking/system.txt
```diff
@@ -0,0 +1,177 @@
+You are an expert rater of student argumentative essays. You rate essays (grades 8-10 student writing) for Critical Thinking, using the rubric and decision rules below. You are one rater among several; rate independently, exactly as a trained human expert would, using ONLY the rubric, the decision rules, the assignment, the source passages, and the essay text. Do not use outside criteria (e.g., general essay quality, spelling, or grammar) except where the rubric or rules make them relevant.
```
adnanrhussain, line 1: @aychi1 - Changed the 6-12 here also to grades 8-10. Are we sure this is 8-10? All configuration indicates 6-12

evals/durable-skills/ela-writing/critical-thinking/system.txt
```diff
@@ -0,0 +1,176 @@
...
+
+
+## Decision rules
+Apply ALL of the following rules.
+
+1. Selection vs. explanation (the copying rule)
+Selecting evidence is a skill, but if selection is all the writer does, that is where the skill stops. Copied or quoted material can satisfy identify/present descriptors: lean Exploring when the writer states evidence, Analyzing when the writer explains what the evidence means. Material restated from the source never satisfies explain/engage descriptors on its own — credit above Exploring needs the evidence tied to the writer's own claim or argument.
+
+2. What counts as a citation
+Citations are defined broadly for grade 8–10 writers: a direct quote, "the article says", an author name, or a source number all count.
```
adnanrhussain, line 110: @aychi1 - Updated this also from 6-12

evals/durable-skills/ela-writing/critical-thinking/fixtures.json
```diff
@@ -0,0 +1,36 @@
...
+    },
+    "expected": {
+      "overall": {
+        "rationale": "Expert consensus median across 6 raters was Analyzing.",
+        "rating": "analyzing"
+      }
+    }
+  },
```
aychi1, line 14: The fixtures should also include the scores for the indicators. Based on this spreadsheet, the indicator scores are all "Analyzing" as well. 
https://docs.google.com/spreadsheets/d/1_sE_Fx2z_bx6l4iKO7PueVYD7jTJR8MR_6y9UgTcmpQ/edit?gid=2100920111#gid=2100920111

evals/durable-skills/ela-writing/critical-thinking/fixtures.json
```diff
@@ -0,0 +1,36 @@
...
+    },
+    "expected": {
+      "overall": {
+        "rationale": "Expert consensus median across 5 raters was Exploring.",
+        "rating": "exploring"
```
aychi1: Similar comment-- the fixture should include the indicators. Based on the consensus_labels_round3, the indicators are all "Exploring", aside from synthesizing_sources which does not have a score.

evals/durable-skills/ela-writing/critical-thinking/fixtures.json
adnanrhussain: P0 - Remove

evals/durable-skills/ela-writing/critical-thinking/config.json
```diff
@@ -0,0 +1,77 @@
...
+  "evaluator": {
+    "id": "durable_skills.ela_writing.critical_thinking",
+    "stable_id": "114fb24f-65d8-4736-bb23-300b0470fc4d",
+    "name": "Critical Thinking Evaluator",
+    "description": "Rates grade 8-10 student argumentative essays for Critical Thinking across five indicators (synthesizing sources, evidence strength, counterarguments, facts over opinions, drawing conclusions) plus an Overall rating, on a five-level scale (not_evident to extending).",
+    "supported_grades": [
+      "8",
+      "9",
+      "10"
+    ]
```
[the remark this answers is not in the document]
adnanrhussain: reconciled to 8-10

evals/durable-skills/ela-writing/critical-thinking/config.json
```diff
@@ -0,0 +1,81 @@
...
+          }
+        }
+      },
+      "model": {
+        "provider": "anthropic",
+        "name": "claude-opus-5"
+      },
+      "generation": {
+        "temperature": null
+      },
```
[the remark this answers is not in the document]
adnanrhussain: already fixed in SDK, and will be validated at implementation time

evals/durable-skills/ela-writing/critical-thinking/fixtures.json
adnanrhussain: @aychi1 - I'll have to include these fields in a followup PR. Nested object verification will require updates to the fixture tests + notebook harness. I dont want to expand the current scope of this PR, and instead do that separately and add later. Is that ok?

evals/durable-skills/ela-writing/critical-thinking/output_schema.json
adnanrhussain: Reviewed with Michelle and made adjustments. Pls see the last commit https://github.com/learning-commons-org/evaluators/pull/188/changes/b484496205d5502bf276d88ffb10a8a287b8fcfd

evals/durable-skills/ela-writing/critical-thinking/output_schema.json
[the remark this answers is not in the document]
adnanrhussain: Fixed

evals/durable-skills/ela-writing/critical-thinking/fixtures.json
aychi1: Yes, sounds good!

aychi1 (approved):
LGTM, thank you so much Adnan for your work on this

PR #208, opened 2026-08-28
refactor(sdk)!: rename TypeScript evaluators onto the eval taxonomy

adnanrhussain:
First of the SDK rebuild PRs. The evaluator contracts under `evals/` are the registry; the SDK still carried its own flat names for the same evaluators. This aligns them and adds the identity fields the contracts already declare.

Mechanical rename plus two real bugs it surfaced. No evaluation behaviour changes — models, temperatures and prompts are untouched.

## Renames

| was | now |
| --- | --- |
| `Conventionality*` | `MeaningDirectness*` |
| `Smk*` | `BackgroundKnowledgeDemands*` |
| `Intertextuality*` | `ReferenceKnowledgeDemands*` |
| `Purpose*` | `PurposeClarity*` |
| `Vocabulary*` | `VocabularyComplexity*` |

Applied to classes, functional wrappers, `*Internal` types, `*OutputSchema` exports, module paths and public exports. `TextComplexityResult` keys follow the evaluators they hold.

## Identity

`metadata.id` is now the dotted registry id read from `config.json`, joined by `stableId` and `idHistory`. Every legacy SDK name is already recorded in its config's `id_history`, so a consumer of an old id can still resolve the evaluator.

`stableId`/`idHistory` are optional on `EvaluatorMetadata` for one reason: `TextComplexityEvaluator` is an SDK-side composite, not a registry evaluator, so it has no definition to read them from.

## Two silent breakages fixed

Both are the same shape — a copy of an id going stale and degrading instead of failing.

- `formatters.ts` pinned `'grade-level-appropriateness'` as a literal, and `report-template.html`'s `EVALUATOR_ORDER` pinned all five. With stale ids `indexOf` misses, GLA stops being recognised, and the report loses its column and its ordering without erroring.
- Two column-prefix sites built names with a raw hyphen replace instead of the existing `formatEvaluatorPrefix` helper, so the new dotted ids would have leaked their namespace into CSV headers. `evaluatorDisplayName` likewise only split on hyphens, which would have rendered `Student_facing_text.ela_reading.meaning_directness` as a report label.

Mutation testing on `formatters.ts` showed why these were missable: the three functions involved had no assertions on them at all. Second commit adds four tests pinning the id → column and id → label derivations, including that one evaluator's score can't land in another's column. The changed region now has a single surviving mutant, an equivalent one — `includes('.')` → `includes("")` is behaviourally identical, since `'sentence-structure'.split('.').pop()` returns the string unchanged.

## Drive-by doc fixes

Two errors in `src/batch/README.md` that predate this PR: the math family's optional column is `grade_level`, not `grade`, and its second credential is the Learning Commons key, not "platform".

## Breaking

Every public class, function and type above; the evaluator ids surfaced in results and telemetry; the batch `--evaluator` member ids; and the per-evaluator CSV column prefixes in batch output.

## Validation

- `tsc --noEmit`, `npm run lint`, `npm run test:unit`, `npm run build`
- `scripts/check.py` from the repo root
- StrykerJS against `src/batch/formatters.ts`, before and after adding the tests
- Reverted each of the three fixes in turn to confirm the new tests actually fail

## Out of scope

- **`demos/typescript`** — pinned to published `0.8.0` and only imports `MathStandardsAlignmentEvaluator`, so nothing breaks. Migrating it later is how we write and validate the published migration guide.
- **Python SDK** — keeps the old names for now, so the two SDKs diverge on naming until it's picked up.
- **`src/schemas/reference-knowledge-demands.ts`** still describes the field as "The Intertextuality complexity level". That file is generated from the eval's `output_schema.json` and the string is model-facing, so it needs an eval-side fix, not a hand-edit here.
- **`README.md` doc links** still point at old doc-site slugs (`/conventionality/`, `/intertextuality`), which live outside this repo.


adnanrhussain:
Reviewed all four (one surfaced, three suppressed).

**The surfaced comment on `text-complexity.test.ts:68` / `:83` is not a defect.** Vitest's `toThrow(string)` matches on substring, not equality, so `"...is required for Text Complexity."` matches the full `"...is required for Text Complexity. Pass googleApiKey in config."` CI was green on both lines.

It was pointing at something real underneath, though: those two assertions stopped at the evaluator name while every sibling test asserts the guidance clause too. Tightened them to the full message, and verified they're load-bearing by reintroducing the ` evaluator` suffix in `base.ts` and confirming both fail.

**Two of the suppressed comments were correct and are now fixed.** `operation: 'smk_evaluation'` and `operation: 'conventionality_evaluation'` were still carrying the old names — my sweep grepped log *messages* and missed the structured `operation:` field. Checked every value across the evaluators: the rest name a config step (`sentence_analysis`, `complexity_classification`, `background_knowledge`, `runCoarseFilter`) and are correct as-is.

PR 2 replaces all of these with config step ids, so this is interim naming rather than a new convention.

Pushed as f6e1e077.

PR #273, opened 2026-09-10
chore: add repo and per-workspace CLAUDE.md guidance

adnanrhussain:
## What

Adds a root `CLAUDE.md` plus three scoped ones so Claude Code (and new contributors) stop re-deriving the repo's layout and verification commands on every session.

| File | Covers |
| --- | --- |
| `CLAUDE.md` | Layout, the `scripts/check.py --fix` gate, per-workspace verify commands, conventional-commits / release-please / generated-file rules |
| `evals/CLAUDE.md` | The `<domain>/<skill-modality>/<evaluator>/` taxonomy, the evaluator file contract, and the `sha256` prompt tripwire |
| `sdks/typescript/CLAUDE.md` | Verify commands, the two published entry points, generated schemas and KG types |
| `demos/typescript/CLAUDE.md` | Consumes the *published* SDK; spec-driven workflow under `specs/` |

## Why

Scoped files only load when work touches that directory, so the per-session cost stays small. Each file points at the existing `README.md` / `CONTRIBUTING.md` / `scripts/README.md` rather than restating them, which keeps them from drifting.

`sdks/python/CLAUDE.md` was dropped per review — Python stays covered at the root level (layout entry, `make verify`, the release-please rule), and a scoped file can follow after 1.0.

Commits use `chore:` rather than `docs:` because `docs` is a visible changelog section for the release-please packages, and agent guidance isn't user-facing package documentation.

## Verification

- `python3 scripts/check.py` — all 6 checks pass
- Every `npm run` script name resolves against the relevant `package.json` (9 in the TS SDK, 2 in the demo)
- Every `make` target resolves against `sdks/python/Makefile` and `evals/Makefile`
- All relative markdown links resolve
- Layout table cross-checked against the tracked top-level dirs — nothing documented that isn't tracked (`docs/` and `tools/` exist only as untracked local dirs, so they're omitted)
- Generated-file claims cross-checked against the `linguist-generated` entries
- The documented `shasum -a 256` command reproduces the `sha256` recorded in `config.json`


evals/CLAUDE.md
```diff
@@ -0,0 +1,46 @@
...
+single-step ones use `system.txt` / `user.txt`, multi-step ones name their own.
+
+## Editing a prompt requires updating its hash
+
+`config.json` pins each prompt file by `sha256`. The `eval-config` check fails
+on drift and **does not auto-fix it** — the hash is a deliberate tripwire, so
+recompute it by hand when you change prompt text:
+
+```shell
+shasum -a 256 <domain>/<skill-modality>/<evaluator>/system.txt
```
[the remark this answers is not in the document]
adnanrhussain: Good catch — fixed in ce82f6c8. The path is now `evals/<domain>/<skill-modality>/<evaluator>/system.txt`, which is consistent with the rest of the guidance (`python3 scripts/check.py` is also root-relative). Verified the command reproduces the `sha256` recorded in `config.json` when run from the repo root.

sdks/python/CLAUDE.md
```diff
@@ -0,0 +1,23 @@
+# sdks/python
```
michael-berger-czi: How about removing for now? I can make a note to add this after 1.0.

michael-berger-czi (approved):
LGTM with an optional question

sdks/python/CLAUDE.md
adnanrhussain: yes, will do that

adnanrhussain: Removed in 1cd4d19c. Python stays covered at the root level (layout entry, `make verify`, the release-please rule), so nothing dangles — it's just the scoped file that's gone.

PR #274, opened 2026-09-10
feat(sdk): deprecate recordInputs and drop opt-in input text telemetry

adnanrhussain:
Telemetry no longer sends the text you evaluate, under any configuration.

## What changed

- `TelemetryEvent.input_text` and the `inputText` argument to `sendTelemetry` are removed. All eight call sites (single-step ×2, multi-step ×2, math standards alignment ×2, vocabulary complexity ×2) stop passing the text; each still passes `textLength`, so `text_length_chars` is unchanged.
- `TelemetryOptions.recordInputs` remains as a `@deprecated` no-op. It is no longer read, and is dropped from `NormalizedTelemetryOptions` and `normalizeTelemetryConfig`.
- New `tests/unit/evaluators/telemetry-input-text.test.ts` covers both halves: setting `recordInputs: true` does nothing, and no event on the success or error path contains the text.
- README's `TelemetryOptions` shape and telemetry section updated to match.

## Verification

`typecheck` clean, `lint` clean, 1386/1386 unit tests pass. Both call shapes above were compiled against this branch to confirm neither errors.

### DIFF
--- a/sdks/typescript/README.md
+++ b/sdks/typescript/README.md
@@ -296,7 +296,7 @@ Every evaluator takes the same options:
 | `modelOverride` | Run every call on a different `{ provider, model }`; `provider` is the exported `Provider` enum |
 | `llmProvider` | Bring your own provider (see below) |
 | `maxRetries` | Retries per failed call (default 2, so 3 attempts) |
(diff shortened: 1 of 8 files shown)


michael-berger-czi (approved):
This LGTM with the exception of the backwards compatibility issue discussed in Slack

PR #276, opened 2026-09-11
refactor(sdk)!: remove the 0.2.0 evaluators, settings machinery, and LangChain provider

michael-berger-czi:
## Summary

Decided to remove the unneeded code first so that the following PRs are clearer.

Claude Summary:
Clears the ground for the Python SDK 1.0 rebuild ([DSCR-2182](https://czi.atlassian.net/browse/DSCR-2182), [plan](https://czi.atlassian.net/wiki/spaces/ETE/pages/4981653510)). Every future phase rewrites the 0.2.0 surface from the shared `evals/` contracts, so this PR deletes it up front rather than adapting it piecemeal. The Phase 1a ticket ([DSCR-2184](https://czi.atlassian.net/browse/DSCR-2184)) lists these deletions; they land here as a standalone PR so the follow-up PRs are purely additive.

**Removed** (~12,300 lines)

- The three 0.2.0 evaluators and their `BaseEvaluator`, input-field wrappers, `answer / explanation` result shape, nested provider configs, `create_config*` factories, `TelemetryConfig`, per-evaluator settings classes, and the old error hierarchy
- The `settings/` package: generated settings modules, bundled `contracts.toml`, `load_settings.py`, `scripts/generate_settings.py`, `scripts/capture.py`
- `providers/langchain.py` and the four `langchain-*` dependencies plus `tomli`
- The recorded-response contract harness and every legacy test
- The shared `sdks/settings/` TOML directory at the repo root
- The docs pages describing that surface (`configuration.md`, `evaluators.md`, `results.md`, `running-evaluations.md`, `error-handling.md`)
- Matching debris in `pyproject.toml`, the Makefile, `.gitattributes`, and the workflows

**Kept**: `logger.py` and its tests, `version.py`, the packaging skeleton, CI. The package now exports only the logger and version until the rebuilt evaluators land; nothing is published from this state.

## Verification

In a fresh venv with only `pydantic` and `textstat` installed: ruff, ruff format check, mypy, `pip check`, and the remaining 10 tests pass on Python 3.13.

## Notes for reviewers

- The recorded LLM responses in the deleted `contracts.toml` files remain at `9124ce4` (this PR's parent) for the shared recorded-response follow-up.
- Stacked: the contract loader / error taxonomy / provider layer PR is based on this branch and retargets to `release-python-sdk-1.0` once this merges.


michael-berger-czi:
Addressing the Copilot review:

- **Local-development references a removed scripts path** — fixed in 25b2dff: the `make lint` line now reads `src/` and `tests/`. The `scripts/` directory (and the line) return in #277 with the contract generator.
- **The release workflow can still publish this state** — accurate about the workflows, which are unchanged, but publishing only happens when a GitHub release is created, and that only happens when someone merges the release-please PR that forms on `main`. This stack targets `release-python-sdk-1.0`, so `main` never carries this state, and the plan deliberately keeps a single release PR to be merged after the last phase rather than gating per phase. Leaving the workflows as they are.

.github/workflows/publish-python-sdk.yml
```diff
@@ -21,11 +21,6 @@
 #    artifacts will exit 0 at publish. Add a step that runs
 #    `test -n "$(ls dist/*.whl 2>/dev/null)"` etc. before publish.
 #
```
adnanrhussain, line 17:
```suggestion
# ─── TODOs / publish-pipeline hardening (tackle in a single follow-up PR) ────
#
# All deferred from the initial 0.1.0 publish PR. Address together once
# sdks-python-v0.1.0 has shipped to PyPI.
#
# 1. Pin `workflow_dispatch` to a release tag. Today it checks out the
#    dispatch ref (typically main HEAD), not the released tag — so a manual
#    retry publishes whatever's currently on main, not the released version.
#    Add `inputs.ref` (required string) to the trigger and pass it to
#    actions/checkout. Or drop workflow_dispatch entirely and retry by
#    deleting + recreating the GitHub Release.
#
# 2. Assert `dist/` is non-empty before the publish step. `skip-existing: true`
#    masks broken builds — a `python -m build` that silently produces zero
#    artifacts will exit 0 at publish. Add a step that runs
#    `test -n "$(ls dist/*.whl 2>/dev/null)"` etc. before publish.
#
```

adnanrhussain (approved):
LGTM. One nit suggestion

PR #277, opened 2026-09-11
feat(sdk): contract loader, spec error taxonomy, and native provider adapters

michael-berger-czi:
## Summary

Phase 1a plumbing. Most of the lines in this PR are generated. Should only be ~2000 lines of non generated, non test code.

AI summary:
Phase 1a PR 1 of [DSCR-2184](https://czi.atlassian.net/browse/DSCR-2184) ([plan](https://czi.atlassian.net/wiki/spaces/ETE/pages/4981653510), Phase 1a). Purely additive on top of the 0.2.0 removal PR; no evaluators yet. PR 2 builds the factories, registry, envelope, and flat config on these interfaces and proves them on the four pilot evaluators.

- **Contract loader and generated modules (D1).** `scripts/generate_contracts.py` bundles every `evals/**` contract (config, input and output schemas, prompt and rubric files) verbatim into `contracts/_generated/` as package data, refusing any prompt whose sha256 drifted from its config, and emits pydantic `<Class>Input` / `<Class>Output` models under `schemas/<family>/<subject>/`. `load_contract(id)` reads a bundled contract into typed objects. `make check-generated` fails CI on stale or orphaned files; the workflow now also triggers on `evals/**`.
- **Error taxonomy (spec §6.1).** `errors.py` transliterates the TS `errors.ts`: abstract `EvaluatorError` / `EvaluationError` / `DependencyError`, retryable as data with the 5xx rule, the closed dependency-id set, and `wrap_provider_error` classifying in TS's fixed order from structured signals only along the cause chain (native SDK status errors, google-genai `APIError.code`, httpx and httpx2 transport errors, pydantic and JSON failures, errnos). Message text never reclassifies.
- **Provider layer (D10).** `LLMProvider` protocol, `Provider` enum, `ProviderConfig`, `create_provider`, and three adapters: OpenAI Responses `parse`, Anthropic `messages.parse`, Google with a raw JSON schema. `call_with_resampling` is the evaluator side of the retry split (resample `LLMOutputProcessingError` only; dependency backoff stays in the vendor SDKs via `max_retries`). Adapters require an explicit key so the vendor SDKs never read the environment (§3.1).

## Decisions worth a look

- Generated input models type `grade_level` as `int` and other inputs as plain `str` / `int` with no bounds or enum constraints, so PR 2's `validate_inputs` is the only validator and messages stay canonical. Accepted tokens are exported as `GRADE_LEVEL_VALUES`, `JURISDICTION_VALUES`, etc.
- Google's pydantic-to-schema conversion rejects the feedback family's integer `enum: [0, 1]`, so the adapter sends `response_json_schema` and validates the returned text itself. The pilot's live Google run in PR 2 is the first real check of that path.
- The Anthropic SDK dropped the `temperature` parameter; the contract's pinned value travels in `extra_body`.
- Math's output model is `MathStandardsAlignmentOutput`, following the uniform `<Class>Output` convention rather than the plan table's `...Result`.
- The bundle includes all 17 contracts, Critical Thinking included; keeping it out of the SDK is the registry conformance allowlist's job in PR 2.

## Verification

Fresh venvs on Python 3.13 and 3.10: `make check-build`, ruff, format check, mypy, `pip check`, and 475 tests pass. The suite includes the spec §6.5 error mapping check against real SDK exception types, adapter tests against fake clients, every bundled contract cross-checked against `evals/` (sha256, placeholder sources, outcome fields), and parser tests with hand-written payloads per `output_schema.json`. A clean wheel contains the 94 bundled contract files and 21 generated schema modules.

| Lines | Count |
| --- | --- |
| Hand-written business logic | ~2,330 |
| Unit tests | ~2,040 |
| Generated schema modules | ~1,120 |

Stacked on the 0.2.0 removal PR; retarget to `release-python-sdk-1.0` once that merges.


sdks/python/pyproject.toml
```diff
@@ -25,6 +25,13 @@ classifiers = [
 dependencies = [
     "pydantic>=2.0.0",
     "textstat>=0.7.0",
+    # Native provider SDKs (D10): one adapter per SDK under providers/. httpx is what
+    # google-genai transports over and what the error taxonomy reads transport failures
+    # from; openai and anthropic bring httpx 2 (the httpx2 package) themselves.
+    "openai>=2.0.0",
+    "anthropic>=1.0.0",
+    "google-genai>=1.30.0",
+    "httpx>=0.27.0",
```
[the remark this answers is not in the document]
michael-berger-czi: Fixed in 097add8, with one correction to the premise: `openai` and `anthropic` do not use the normal `httpx` package here. Both transport over httpx 2, published as the separate `httpx2` package (`openai` since 2.47.0, `anthropic` since 1.0.0), and their exception classes are built on `httpx2.Request` / `httpx2.Response`, which is what the tests construct.

The real gap was the lower bound: `openai>=2.0.0` admitted 2.0–2.46, which depend on `httpx` and so leave `httpx2` uninstalled. The bound is now `openai>=2.47.0` (the first release on httpx2), so a `pip install -e ".[dev]"` always brings `httpx2`. `errors.py` was already tolerant of either generation via `sys.modules` lookups; the comment in `pyproject.toml` now explains the dependency.

sdks/python/src/learning_commons_evaluators/contracts/loader.py
```diff
@@ -0,0 +1,348 @@
...
+    values: list[str] = Field(alias="in")
+
+    @field_validator("values", mode="before")
+    @classmethod
+    def _as_strings(cls, raw: Any) -> Any:
+        # The schema admits numbers; inputs are compared as the strings the caller passes.
+        return [str(v) for v in raw] if isinstance(raw, list) else raw
+
+    def holds(self, fields: dict[str, str]) -> bool:
+        return fields.get(self.input) in self.values
```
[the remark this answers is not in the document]
michael-berger-czi: Comment from Claude: Applied in 64ef417 and repaired in 0550b98. The applied suggestion kept the old `def holds` line above the new one, so the module did not parse (CI caught it). `holds` now takes a read-only `Mapping[str, object]`, compares as strings, and never holds for a missing input; covered for an int grade, a missing key, and `None`.

sdks/python/src/learning_commons_evaluators/providers/retry.py
```diff
@@ -0,0 +1,65 @@
...
+    DependencyId,
+    EvaluatorError,
+    LLMOutputProcessingError,
+    wrap_provider_error,
+)
+
+T = TypeVar("T")
+
+
+async def call_with_resampling(
```
adnanrhussain, line 27: P2 - dependency backoff is delegated entirely to the vendor SDKs. not an issue here, but will be applicable in post-v1 where we inject llm providers. Also, the KG client may need this. Ok to review and handle if needed in a downstream PR

sdks/python/pyproject.toml
```diff
@@ -46,7 +56,15 @@ Documentation = "https://docs.learningcommons.org/evaluators"
...
 [tool.setuptools.package-data]
-learning_commons_evaluators = ["py.typed"]
+learning_commons_evaluators = [
+    "py.typed",
+    # The evals/ registry as the package ships it: config.json, input and output
+    # schemas, and the prompt files each contract names. Written by
+    # `make generate-contracts` (scripts/generate_contracts.py) and read by
+    # contracts/loader.py; the canonical copy is evals/ at the repo root.
+    "contracts/_generated/**/*.json",
+    "contracts/_generated/**/*.txt",
```
adnanrhussain, line 66: P2 - ok for now. we should add a CI check to ensure all assets are either .txt / .json to ensure any other format does not pass-through silenlty until runtime failures

sdks/python/pyproject.toml
```diff
@@ -25,6 +25,16 @@ classifiers = [
...
     "textstat>=0.7.0",
+    # Native provider SDKs (D10): one adapter per SDK under providers/. httpx is what
+    # google-genai transports over and what the error taxonomy reads transport failures
+    # from. openai (since 2.47.0) and anthropic (since 1.0.0) transport over httpx 2,
+    # published as the separate httpx2 package, which they bring themselves; the lower
+    # bounds are the first versions to do so, so httpx2 is always present for the error
+    # classifier and the tests that build the SDKs' exceptions.
+    "openai>=2.47.0",
+    "anthropic>=1.0.0",
+    "google-genai>=1.30.0",
```
adnanrhussain, line 36: P1 - should we set an upper bound, on whats this tested on? And dependabot will bump this up?

sdks/python/src/learning_commons_evaluators/errors.py
```diff
@@ -0,0 +1,672 @@
...
+    current: BaseException | None = error
+    depth = 0
+    while current is not None and depth < _MAX_CHAIN_DEPTH and id(current) not in seen:
+        seen.add(id(current))
+        yield current
+        depth += 1
+        if current.__cause__ is not None:
+            current = current.__cause__
+        elif not current.__suppress_context__:
+            current = current.__context__
```
adnanrhussain, line 315: _(Reviewed by Claude Code.)_

**Issue:** walking `__context__` lets an unrelated exception decide the class. `__context__` is "what was being handled when this was raised", not a cause — so a bug of ours raised inside an `except httpx.TimeoutException:` block gets classified as a retryable `RequestTimeoutError` and resampled against a provider that is fine. It also makes the class depend on whether the raising code happened to write `from None`, which isn't a signal anyone sets deliberately. TS's `causeChain` follows `cause` only, and SPEC §6.5 says "walking the cause chain".

**Suggested fix:** drop the `__context__` branch, so the walk only follows explicit `raise ... from`:

```python
current = current.__cause__
```

**To verify:**

```python
try:
    try:
        raise httpx.ReadTimeout("upstream slow")
    except httpx.TimeoutException:
        raise ValueError("bug in our own post-processing")   # no `from`
except Exception as e:
    print(type(wrap_provider_error(e, dependency="google")).__name__)
```

Prints `RequestTimeoutError` (retryable) today, and `LLMProviderError` (not retryable) if you add `from None` to that raise. Both should be the catch-all. Worth a test pinning it, since the existing `wrap_provider_error` tests all build the chain with an explicit cause and so never exercise this path.

adnanrhussain (approved):
LGTM 🚀

sdks/python/pyproject.toml
michael-berger-czi: Comment from Claude: Done in 30cd202. Every runtime dependency is now capped at its next major (`openai>=2.47.0,<4`, `anthropic>=1.0.0,<2`, `google-genai>=1.30.0,<3`, `httpx>=0.27.0,<1`, plus `pydantic<3` and `textstat<1`), with a comment recording the versions the suite was verified against (openai 3.13, anthropic 1.5, google-genai 2.23, httpx 0.28; a fresh resolve today also passes on openai 3.14 and anthropic 1.6). The reasoning: the adapters and the error classifier read vendor-specific surfaces (`responses.parse`, `messages.parse` + `extra_body`, `response_json_schema`, exception attributes) that a major can change, so a new major should be admitted by a deliberate bump after the suite has run against it, not by resolution.

On Dependabot: it was not configured for Python at all (npm and github-actions only), so nothing would have bumped these. Added a `pip` entry for `/sdks/python` grouped like the npm one (minor/patch grouped weekly; a major shows up as its own PR against the cap, which is the moment to run the suite against it).

sdks/python/src/learning_commons_evaluators/errors.py
michael-berger-czi: Comment from Claude: Agreed and fixed in c2c7e48. `_cause_chain` now follows `__cause__` only, matching TS's `causeChain` and §6.5; the docstring explains why `__context__` is deliberately not followed. Tests pin both cases from your snippet: a `ValueError` raised while handling an `httpx.ReadTimeout` (no `from`) and the `from None` variant both classify as a non-retryable `LLMProviderError`, while an explicit `raise ... from upstream` still resolves the 429 beneath it. The earlier test that asserted the opposite behaviour was replaced.

PR #278, opened 2026-09-11
feat(sdk): single-step evaluators, flat config, envelope, and registry with three pilots

michael-berger-czi:
## Summary

Single step implementation for a select few evaluators to verify plumbing.

Testing notes: I've created a git workstream off of main so I can verify the before-and-after project shapes at the same time.

`git worktree add --detach ../2.0-python-view origin/main`

AI summary:

Phase 1a PR 2 of [DSCR-2184](https://czi.atlassian.net/browse/DSCR-2184), re-scoped per the [review comment](https://czi.atlassian.net/browse/DSCR-2184?focusedCommentId=989088) to single-step only: Grade Level Appropriateness stays, Vocabulary Complexity (multi-step) moves to the next PR, Purpose Clarity replaces it, and Tone Appropriateness is the one feedback pilot. Stacked on #277 (contract loader, errors, providers); retarget to `release-python-sdk-1.0` when that merges.

The three pilots cover the shapes the later single-step ports need: text-only and grade-free with a grade-band output (GLA, Google), text + grade with contract-declared preprocessing (Purpose Clarity, Google, `textstat` Flesch-Kincaid bound to `{fk_score}`), and the feedback family's two-text input with a binary `quality_score` (Tone, OpenAI). Every later single-step evaluator is a three-line declaration onto these interfaces.

## What's in it

- **`EvaluatorConfig`** (spec §3): provider keys, `learning_commons_api_key`, `model_override`, `max_retries`, `telemetry`. Pass it whole or as evaluator kwargs: `PurposeClarityEvaluator(google_api_key="...")`.
- **`BaseEvaluator`**: required keys derived from the contract (step providers + `required_credentials`); `model_override` replaces the provider half only, is validated, and is warned once (§3.2). Canonical `Missing required credential: {key}` before any I/O.
- **`SingleStepEvaluator`**: the Python form of TS's `defineSingleStepEvaluator`. A concrete evaluator names its contract and generated models; the class resolves the `evaluate_{slug}` step, provider, preprocessing, and metadata at import, so a contract it cannot run fails at import. Flow: validate (§4.1 order and messages, `grade_level: int` rendered as its token), preprocess, render (plain `{name}` substitution, identical to TS), call with immediate resampling of `LLMOutputProcessingError`, envelope.
- **Envelope and `read_outcome`** (§5.1, D2): `EvaluationResult(evaluator, result, metadata)` with `metadata.model` as the `provider:model` that ran; `read_outcome` reads the contract's `outcome` block, `None` score when absent.
- **Registry**: `get_evaluators()`, `get_evaluator(id)` resolving `id_history`, collision guard.
- **Docs**: README quick start and configuration table; local-development guide gains "Adding an evaluator".

## Verification

- `tests/unit/evaluators/test_registry_conformance.py`: every `evals/**/config.json` has a registered class or an entry in `UNIMPLEMENTED` (14 entries, self-cleaning: porting one without deleting its entry fails), identity/grades/providers/credentials/outcome match the contract, modules sit at the id-derived path, temperature sent matches, every declared placeholder reaches the prompt, minimum lengths enforced.
- `tests/unit/test_cross_sdk_prompts.py`: for all 19 fixtures, Python renders what the TS renderer (`replaceAll` over declared keys) renders, with `fk_score` masked.
- Single-step flow on a synthetic contract (refusals, contract reading, inputs, envelope, error attribution, resampling, log hygiene), config and override checks, input validation wording, registry, `read_outcome`, preprocessing dispatch.
- **Live** (`make integration-test`, gated on `RUN_INTEGRATION_TESTS=1` + keys): all 19 fixtures run against the real providers. 18 match exactly on the first attempt. The GLA `trees` case returns `2-3` consistently where the fixture expects `4-5`; the TS live test applies the same tolerance (three attempts, then adjacent accepted), so it passes there too. Prompt rendering is byte-identical, so this is model judgment at `temperature: 1`, not a port defect.
- Full gate green on Python 3.13 and 3.10: ruff, format, mypy, `check-generated`, 646 unit tests (+19 integration skipped without keys).

## Decisions worth a look

- Declarative subclassing (`class X(SingleStepEvaluator[In, Out]): contract = ...`) instead of a factory function: same load-time resolution as TS, and mypy can type the result.
- `evaluate()` accepts the typed input model or keyword fields, not both. Argument-shape misuse is a `TypeError`; everything from validation on is an `EvaluatorError`.
- Computed values render like JavaScript's `String(number)` (`7` not `7.0`), so the two SDKs bind identical text.
- Telemetry options are accepted (so the config surface is final) but nothing is emitted yet; that is the telemetry PR.


sdks/python/src/learning_commons_evaluators/features/preprocessing.py
```diff
@@ -0,0 +1,72 @@
...
+    return float(fn(text))
+
+
+_LIBRARY_ADAPTERS: dict[str, Callable[[str, str], float]] = {
+    "textstat": _textstat,
+}
+
+
+def _round(value: float, transform: PostTransform) -> float:
+    return round(value, transform.precision or 0)
```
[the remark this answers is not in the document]
michael-berger-czi: Fixed in dc94413: the `round` transform now computes `floor(value * factor + 0.5) / factor`, the JavaScript `Math.round` rule, with tie cases (`0.125 → 0.13`, `2.675 → 2.68`, `-2.675 → -2.67`) covered in `tests/unit/features/test_preprocessing.py`. One case documents that `1.005` lands on `1.0` in both languages, since the binary product is `100.4999…`, so the two SDKs agree there as well.

sdks/python/tests/unit/evaluators/feedback/ela_writing/test_tone_appropriateness.py
```diff
@@ -0,0 +1,64 @@
...
+    evaluator = ToneAppropriatenessEvaluator(openai_api_key="k")
+    await evaluator.evaluate(ToneAppropriatenessInput(student_text=STUDENT, feedback_text=FEEDBACK))
+    [system, user] = providers.last.calls[0]["messages"]
+    template = evaluator.contract.document("user.txt")
+    assert user["content"] == template.replace("{student_text}", STUDENT).replace(
+        "{feedback_text}", FEEDBACK
+    )
+    assert (
+        "{" not in user["content"].replace("{student_text}", "").replace("{feedback_text}", "")
+        or True
```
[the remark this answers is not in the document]
michael-berger-czi: Fixed in dc94413: the test now asserts directly that neither `{student_text}` nor `{feedback_text}` remains in the rendered user prompt. The `or True` was a scratch line that should not have been committed.

michael-berger-czi:
End to end tests so far: https://czi.atlassian.net/browse/DSCR-2196?focusedCommentId=990131

sdks/python/src/learning_commons_evaluators/features/preprocessing.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+def run_preprocessing_step(text: str, implementation: Implementation) -> float:
+    """Run a single library computation and its declared post-transform.
+
+    :raises ValueError: for a library or transform type this SDK does not provide.
+    """
+    adapter = _LIBRARY_ADAPTERS.get(implementation.library)
+    if adapter is None:
+        raise ValueError(
```
adnanrhussain, line 53: P1 - Pls confirm, I think the single_step is raising this as an LLMProviderError, for an issue caused by us on a bad processing library configuration. 
Maybe this should be surfaced as a ConfigurationError, or we should introduce a new error NotSupported error class?

sdks/python/src/learning_commons_evaluators/prompts/render.py
```diff
@@ -0,0 +1,22 @@
...
+from __future__ import annotations
+
+from collections.abc import Iterable, Mapping
+
+
+def render_prompt(template: str, inputs: Mapping[str, str], placeholders: Iterable[str]) -> str:
+    text = template
+    for name in placeholders:
+        if name in inputs:
+            text = text.replace(f"{{{name}}}", inputs[name])
```
adnanrhussain, line 18: is this replaceAll? I am not very confident, but I think there was one place with multiple placeholders

adnanrhussain (approved):
lgtm 🚀

sdks/python/src/learning_commons_evaluators/prompts/render.py
michael-berger-czi: Comment from Claude: Yes, this is replace-all. Python's `str.replace(old, new)` with no `count` argument replaces every occurrence, so it is the same operation as TS's `text.replaceAll(`{${key}}`, inputs[key])` in `createPromptRenderers`. Checked against the registry: the one prompt with a repeated placeholder is `sentence-structure/complexity-user.txt` (`{grade_level}` twice), and `render_prompt("{a} and {a} and {b}", {"a": "X"}, ["a"])` gives `X and X and {b}`: both occurrences substituted, the undeclared `{b}` left alone, as in TS. The cross-SDK prompt check in `tests/unit/test_cross_sdk_prompts.py` re-states the TS algorithm independently and compares every fixture's rendering byte for byte, so a divergence here would fail that suite.

sdks/python/src/learning_commons_evaluators/features/preprocessing.py
michael-berger-czi: Comment from Claude: Confirmed, and fixed in 1202f27. A contract naming a library, function or transform this SDK lacks passed class creation and only failed inside `evaluate()`, where the error boundary wrapped it as `LLMProviderError` with `dependency=google` (reproduced before the fix). The TS single-step factory has the same gap.

Rather than a new taxonomy leaf, the check moved to class creation: `SingleStepEvaluator.__init_subclass__` now calls `check_implementation` on every preprocessing entry and raises `NotImplementedError` naming the evaluator and the missing piece (`Preprocessing "fk_score" in Thing Evaluator config.json: Unsupported preprocessing library "nltk". Supported: textstat.`), so an unrunnable contract fails at import like the other contract-shape guards and can never be misattributed to a provider. `run_preprocessing_step` raises the same for direct callers. Three tests pin it (library, function, transform).

Why not a `NotSupportedError` in the taxonomy: §6.1 requires a new class to be added to the spec and implemented in TS too, and this is a build-time/registry problem rather than a runtime fault domain the caller handles; Python's built-in `NotImplementedError` is the idiomatic signal for "this implementation does not provide it". If you still want a public class for it, that is a small follow-up once the spec has it.

PR #283, opened 2026-09-15
ci: add dependabot version updates and a verification gate for the TypeScript demo

adnanrhussain:
## What

Brings `demos/typescript` under Dependabot version updates, and adds the CI job that verifies those bumps.

- **`.github/dependabot.yml`** — new `npm` entry for `/demos/typescript`. Same weekly Monday 10:00 PT cadence and `dependencies`/`npm` labels as the SDK entry, limit 5, but a single `demo-dependencies` group matching `*` across major/minor/patch.
- **`.github/workflows/test-demo-typescript.yml`** — path-filtered on `demos/typescript/**`, runs `npm run typecheck` then `npm run build` on Node 22. Reuses the SDK workflow's `actions/checkout@v7` + `bahmutov/npm-install@v1` conventions.
- **`demos/typescript/package.json`** — adds `"build": "vite build"`.

## Why

The demo had no version-update entry, so it only ever received Dependabot *security* updates (which ignore `dependabot.yml`): #146, #154, #267, #269, #272, #275. Those six landed with **no CI verifying them** — `checks.yml` is Python-only and `test-sdk-typescript.yml` is scoped to `sdks/typescript/**`.

The demo is `private: true` and unpublished, so the SDK's caution around major bumps buys nothing here — grouping everything into one weekly PR is cheaper to review. That grouping is only defensible because CI now gates it, hence the two changes landing together.

The `build` script is new because the demo had none. `typecheck` alone would green-light a bump that breaks Vite plugin or config resolution (e.g. a `@vitejs/plugin-react` major), which is precisely the failure mode a grouped-majors PR needs to catch. `dist/` is already gitignored.

## Verification

`npm ci && npm run typecheck && npm run build` pass on this branch (Node 22.23.2, 117 modules transformed).


michael-berger-czi (approved):
LGTM

czi-fsisenda (approved):
LGTM!

PR #284, opened 2026-09-15
feat(sdk): Vocabulary Complexity and Sentence Structure on the multi-step base

michael-berger-czi:
Summary per Claude: this is the second of two PRs for Phase 1b ([DSCR-2213](https://czi.atlassian.net/browse/DSCR-2213)) — the two contracts that need more than one model call, declared on the base in #286. **Base is #286's branch**, so review that one first; retarget this to `release-python-sdk-1.0` once it merges. Both evaluators are declarations plus the computations their contracts name; nothing about the multi-step flow lives here.

> Restructured from the original single PR at the reviewer's request. The plumbing moved to #286, so the two earlier Copilot threads on `multi_step.py` now point at lines that live there. Both were addressed before the split — the provider cache key and `as_prompt_text`'s JSON parity — and the fixes are in #286.

## The two evaluators

- **Vocabulary Complexity** (45 lines): a shared prose step for background knowledge, then the rating branch the grade selects — Gemini 2.5 Pro for grades 3-4 with `{fk_score}` bound, GPT-4.1 for 5-12 without it, as each step's and each preprocessing entry's own condition declares. It is short because the routing is the contract's, not the code's.
- **Sentence Structure**: `textstat` ground-truth counts into an LLM grammatical analysis, then ~30 ratios derived from its counts into the grade's rubric for classification. `SentenceAnalysis` states the first step's shape, which the registry does not describe — `config.json` declares an `output_schema` for the evaluator, not one per step.

`features/readability.py` (the ground-truth counts block) and `features/sentence_features.py` (the ratios and their JSON form) hold the computations the contracts name as `custom`, keyed by the declared function name so a rename fails at import. The derived features are byte-identical to the internal implementation the fixtures came from over 5,000 randomised analyses.

## Verification

- Registry conformance and the cross-SDK prompt check now **derive** the steps an evaluator runs from the contract and compare every one of them, instead of assuming a single conventionally-named step. Placeholders each SDK computes with its own language's library are masked, and so are step outputs, which would mean calling a model; the masked numbers' rounding is checked separately.
- The integration suite derives the expected envelope label from the contract per case — every model the steps *these inputs* run will use, deduplicated, in run order — and asserts it exactly.
- 785 tests on Python 3.13 and 3.10; all 22 live fixtures pass, both new evaluators exact on the first attempt.

## Known cross-SDK divergences, flagged in code

Tracked in [DSCR-2223](https://czi.atlassian.net/browse/DSCR-2223), not fixed here, because both touch published TypeScript 1.0 behaviour:

1. `features_to_json` truncates each feature to a whole number — the notebook's arithmetic, which produced the fixtures and rubric thresholds — where TypeScript `Math.round`s. Measured over 2,000 synthetic analyses, every one differs on at least one feature.
2. TypeScript clamps the ground-truth Flesch-Kincaid grade at zero; Python does not, so very simple text can bind a negative grade.

The registry pins neither, because `kind: "custom"` means the contract *names* a function rather than describing it — which is how the two drifted with no check failing.

[DSCR-2224](https://czi.atlassian.net/browse/DSCR-2224) separately tracks that a contract fault at evaluation time is classified as a provider failure, in both bases.


adnanrhussain (approved):
LGTM

czi-fsisenda (commented):
Looks good!

PR #285, opened 2026-09-15
ci(sdk-python): gate Python SDK coverage at 95% and isolate Codecov per SDK

adnanrhussain:
## Summary

The Python SDK suite sits at **98.06%** line coverage (660 passed, 19 skipped), but CI neither enforced nor reported it: `make verify` doesn't run coverage, `make coverage` had no floor, and nothing uploaded to Codecov. The TypeScript SDK already has a coverage job and a Codecov upload, so the gap was one-sided.

This adds a Python coverage gate and the Codecov configuration needed for the two SDKs to be judged independently.

## Changes

**`sdks/python/Makefile`** — `COV_MIN ?= 95`; `make coverage` now writes `coverage.xml` and exits non-zero below the floor. The 3-point gap under the measured 98.06% is headroom so one uncovered line doesn't block an otherwise good PR.

**`.github/workflows/test-sdk-python.yml`** — new `Coverage` job on 3.13 running `make coverage`, then uploading under `flags: python-sdk`. The `make` step is the gate; the upload is `continue-on-error` so a Codecov outage can't redden a passing suite, and `if: always()` so a coverage drop still gets its per-line annotation.

**`codecov.yml`** (new) — per-flag isolation, validated against `codecov.io/validate`:
- `carryforward: true` per flag. Both SDK workflows are path-filtered, so a Python-only PR never runs the TS coverage job. Without carryforward, Codecov reads the absent TS coverage as coverage *lost* and posts a large phantom drop on every Python PR.
- `project.default: false` / `patch.default: false` — no repo-wide aggregate, so a TypeScript gap can never fail a Python PR.
- `typescript-sdk` is `informational: true`, reporting without blocking until that suite has a floor of its own.
- `require_ci_to_pass: false` — on the default, a failing TS job would withhold the Python status and stall the PR on an unrelated suite.
- `fixes: ["src/::sdks/python/src/"]` — coverage.py names files relative to the directory `make` ran in (`src/learning_commons_evaluators/...`) while the upload step runs at the repo root. Without the rewrite every `paths` filter matches zero files and the status is computed on an empty report.

**`.github/workflows/test-sdk-typescript.yml`** — adds `codecov.yml` to its path filters, matching the Python workflow, so a PR editing only the coverage config still re-runs the TS upload rather than having its status judged against a carried-forward report from an unrelated commit.

**`sdks/python/docs/local-development.md`** — notes that `make coverage` now fails below the floor.

## Verification

- `make verify` — 660 passed, 19 skipped
- `make coverage` — passes at 98.06% against the 95% floor
- `make coverage COV_MIN=99` — fails as intended (`Required test coverage of 99% not reached`)
- `python scripts/check.py` — all 6 repo checks pass
- `codecov.yml` — "Valid!"; validator confirms `fixes` anchors to `^src/`, `project.python-sdk` target 95.0 with no threshold, `project.default: false`

## Notes

Base is `release-python-sdk-1.0`, not `main` — the Python SDK diverges by 252 files between them, so a gate written against `main` would target a different codebase.

No branch-protection changes here. Neither SDK test workflow is a required check today; making `codecov/project/python-sdk` required is a separate, manual step, deliberately left to the repo owner.

Known caveat: on a fork PR (`CODECOV_TOKEN` unavailable) or a Codecov outage, the upload fails silently and carryforward lets the *patch* status be computed from the previous commit's report. The `make coverage` line floor still holds deterministically; patch coverage has no Makefile equivalent, so treat it as advisory until the status is made required.


michael-berger-czi (approved):
LGTM

PR #286, opened 2026-09-15
feat(sdk): the multi-step evaluator base

michael-berger-czi:
Summary per Claude: this is the first of two PRs for Phase 1b ([DSCR-2213](https://czi.atlassian.net/browse/DSCR-2213)) — the flow for a contract with more than one model call, with no evaluator on it yet. Split out of #284 so the plumbing is reviewable on its own; #284 now stacks on this branch and carries the two pilots that use it (Vocabulary Complexity, Sentence Structure). Same shape as #277, which shipped the loader, errors and providers before any evaluator consumed them.

## What it does

Steps run in the order `config.json` declares them, skipping any whose `condition` the inputs miss, so the branch a grade takes and the fact that one step reads another are consequences of the contract rather than of hand-written control flow. Conditional *steps* are handled here, which the TypeScript `defineMultiStepEvaluator` does not do — that is why Vocabulary Complexity is hand-written there and can be declared here.

A subclass names its contract, its generated models, a `step_models` entry for **every** declared step (`None` for one that answers in prose, called as text and trimmed), and `computations` keyed by `implementation.python.function` for `custom` preprocessing entries. Everything else is read from the contract at class creation, so a contract that gains, renames or re-pins a step fails at import rather than drifting.

- Placeholders resolve from the four sources a contract can name: a caller input, `input.<field>`, an earlier step's output, and a preprocessing output.
- Preprocessing is selected by output **name** rather than entry id, so several entries can declare the same output under mutually exclusive conditions. Each output is computed once per evaluation however many placeholders read it.
- One client per distinct provider and model, keyed on the model that will actually be called, so a `model_override` collapses every step onto one client instead of building an identical one per declared step.
- The envelope sums token usage across the steps that ran and names every model that ran, joined with `+`. A failure is attributed to the step that was running, not the last one, so a contract spanning two vendors never names a service it did not call.
- `as_prompt_text` renders a structured step output as JavaScript's `JSON.stringify` does: compact separators, the characters themselves rather than `\uXXXX` escapes, an integral float as `2`, and `null` for a non-finite one.

## Shared with the single-step base

`EvaluatorMetadata.from_contract` and its `label`, `BaseEvaluator._raw_fields`, and `BaseEvaluator.effective_model` — so both bases read the contract's static facts, accept the caller's two input forms, and resolve a `model_override` one way. `features.round_half_up` is now exported as the SDK's one rounding operation, which `_round` (the contract's `post_transform`) delegates to.

## Verification

714 tests on Python 3.13 and 3.10. `tests/unit/evaluators/test_multi_step.py` drives synthetic contracts that reach every branch: conditional steps and conditional entries, a prose step feeding a structured one, a computation reading a step's output, forward references, per-step failure attribution and resampling, the envelope's joined label under and without an override, and each refusal at class creation (a single-step contract, a step with no `step_models` entry, a `step_models` entry the contract does not declare, a last step that does not produce the output model, an optional step, an unreadable placeholder source, a placeholder naming an undeclared step, a `custom` function the evaluator does not supply, and an unbundled rubric).


adnanrhussain (approved):
LGTM. 

@michael-berger-czi  - As you already shared, this work surfaced TS SDK gaps also. I dug a lil deeper on tests, and used mutation tests to see where more such gaps are. Here are some P1 findings as tests, I put for your review here - https://github.com/learning-commons-org/evaluators/blob/ahussain/pr286-p1-findings/sdks/python/tests/unit/evaluators/test_multi_step_p1_findings.py
Each test is self-documented and some should be kept for coverage even after the fix

Once this stack merges, I will work to put up some shared config contract tests that work across SDKs.

michael-berger-czi:
Thanks for the review @adnanrhussain! To clarify, are you proposing adding those new tests to the top of this stack?

adnanrhussain:
@michael-berger-czi - None of these are P0s, but maybe foundational fixes / tests to include moving forward.
My request for now, if for you to review the file of tests I linked. If you feel the test identifies a valid issue, use that test as reference to fix the code. The file includes comments with suggestions on which tests to keep (ex: `**On fixing:** keep this test`). If you decide to use this, I would recommend including in 286, or in another PR in the stack.

michael-berger-czi:
Thanks @adnanrhussain, I fixed two issues that you found, and made a ticket for the registry issues: https://czi.atlassian.net/browse/DSCR-2228 and https://czi.atlassian.net/browse/DSCR-2227. I'll be doing some more confirmations in the meantime.

PR #289, opened 2026-09-16
fix(docs): Reduce bloat in main README + Keep details in evaluator READMEs

esthersweon:
- Renamed "Strength Acknowledgement" => "Strength Acknowledgment"
- Kept links to evaluator-specific notebooks, prompts, etc. in evaluator READMEs only


adnanrhussain (approved):
🚀

PR #291, opened 2026-09-17
feat(sdk): Python SDK telemetry emission from the evaluator base

michael-berger-czi:
Adding telemetry events

```
{
	"timestamp": "2026-09-17T22:13:15.584Z",
	"sdk_version": "learning-commons-evaluators-python-0.2.0",
	"evaluator_type": "feedback.ela_writing.tone_appropriateness",
	"grade": "",
	"status": "success",
	"latency_ms": 7934,
	"text_length_chars": 154,
	"provider": "openai:gpt-5.4-2026-03-05",
	"token_usage": {
		"input_tokens": 1086,
		"output_tokens": 379
	},
	"metadata": {
		"stage_details": [{
			"stage": "evaluate_tone_appropriateness",
			"provider": "openai:gpt-5.4-2026-03-05",
			"latency_ms": 7934,
			"token_usage": {
				"input_tokens": 1086,
				"output_tokens": 379
			}
		}]
	}
}

{
	"timestamp": "2026-09-17T22:13:34.644Z",
	"sdk_version": "learning-commons-evaluators-python-0.2.0",
	"evaluator_type": "student_facing_text.ela_reading.grade_level_appropriateness",
	"grade": "",
	"status": "success",
	"latency_ms": 16736,
	"text_length_chars": 441,
	"provider": "google:gemini-3.6-flash",
	"token_usage": {
		"input_tokens": 1369,
		"output_tokens": 322
	},
	"metadata": {
		"stage_details": [{
			"stage": "evaluate_grade_level_appropriateness",
			"provider": "google:gemini-3.6-flash",
			"latency_ms": 16734,
			"token_usage": {
				"input_tokens": 1369,
				"output_tokens": 322
			}
		}]
	}
}
{
	"timestamp": "2026-09-17T22:13:43.400Z",
	"sdk_version": "learning-commons-evaluators-python-0.2.0",
	"evaluator_type": "student_facing_text.ela_reading.purpose_clarity",
	"grade": "3",
	"status": "success",
	"latency_ms": 8229,
	"text_length_chars": 885,
	"provider": "google:gemini-3-flash-preview",
	"token_usage": {
		"input_tokens": 1183,
		"output_tokens": 592
	},
	"metadata": {
		"stage_details": [{
			"stage": "evaluate_purpose_clarity",
			"provider": "google:gemini-3-flash-preview",
			"latency_ms": 7921,
			"token_usage": {
				"input_tokens": 1183,
				"output_tokens": 592
			}
		}]
	}
}
{
	"timestamp": "2026-09-17T22:13:50.184Z",
	"sdk_version": "learning-commons-evaluators-python-0.2.0",
	"evaluator_type": "student_facing_text.ela_reading.sentence_structure",
	"grade": "3",
	"status": "success",
	"latency_ms": 6255,
	"text_length_chars": 93,
	"provider": "openai:gpt-4o-2024-08-06",
	"token_usage": {
		"input_tokens": 3643,
		"output_tokens": 548
	},
	"metadata": {
		"stage_details": [{
			"stage": "sentence_analysis",
			"provider": "openai:gpt-4o-2024-08-06",
			"latency_ms": 4494,
			"token_usage": {
				"input_tokens": 2062,
				"output_tokens": 418
			}
		}, {
			"stage": "classify_complexity",
			"provider": "openai:gpt-4o-2024-08-06",
			"latency_ms": 1758,
			"token_usage": {
				"input_tokens": 1581,
				"output_tokens": 130
			}
		}]
	}
}
{
	"timestamp": "2026-09-17T22:14:11.784Z",
	"sdk_version": "learning-commons-evaluators-python-0.2.0",
	"evaluator_type": "student_facing_text.ela_reading.vocabulary_complexity",
	"grade": "3",
	"status": "success",
	"latency_ms": 21023,
	"text_length_chars": 1058,
	"provider": "google:gemini-2.5-pro",
	"token_usage": {
		"input_tokens": 4564,
		"output_tokens": 351
	},
	"metadata": {
		"stage_details": [{
			"stage": "background_knowledge",
			"provider": "openai:gpt-4o-2024-11-20",
			"latency_ms": 1372,
			"token_usage": {
				"input_tokens": 2698,
				"output_tokens": 63
			}
		}, {
			"stage": "vocab_complexity_grades_3_4",
			"provider": "google:gemini-2.5-pro",
			"latency_ms": 19645,
			"token_usage": {
				"input_tokens": 1866,
				"output_tokens": 288
			}
		}]
	}
}
{
	"timestamp": "2026-09-17T22:14:25.894Z",
	"sdk_version": "learning-commons-evaluators-python-0.2.0",
	"evaluator_type": "feedback.ela_writing.tone_appropriateness",
	"grade": "",
	"status": "success",
	"latency_ms": 13008,
	"text_length_chars": 154,
	"provider": "anthropic:claude-opus-5",
	"token_usage": {
		"input_tokens": 1837,
		"output_tokens": 976
	},
	"metadata": {
		"stage_details": [{
			"stage": "evaluate_tone_appropriateness",
			"provider": "anthropic:claude-opus-5",
			"latency_ms": 13008,
			"token_usage": {
				"input_tokens": 1837,
				"output_tokens": 976
			}
		}]
	},
	"model_override": true
}


```

Model override events were also checked manually using Charles Proxy.
```
{
	"timestamp": "2026-09-17T22:28:58.646Z",
	"sdk_version": "learning-commons-evaluators-python-0.2.0",
	"evaluator_type": "student_facing_text.ela_reading.purpose_clarity",
	"grade": "3",
	"status": "success",
	"latency_ms": 30724,
	"text_length_chars": 885,
	"provider": "anthropic:claude-opus-5",
	"token_usage": {
		"input_tokens": 3432,
		"output_tokens": 2138
	},
	"metadata": {
		"stage_details": [{
			"stage": "evaluate_purpose_clarity",
			"provider": "anthropic:claude-opus-5",
			"latency_ms": 30509,
			"token_usage": {
				"input_tokens": 3432,
				"output_tokens": 2138
			}
		}]
	},
	"model_override": true
}
```
Claude summary:
Phase 1c (DSCR-2186). 0.2.0 accepted a telemetry config and sent nothing — `send_telemetry` was a TODO — so this is the emit path, and it lives in `BaseEvaluator`: an evaluator reports by running its evaluation inside `_telemetry_run`, which both bases do, so the five registered evaluators and every evaluator the bases build later emit without code of their own. One event per evaluation, on success and on failure, input validation included.

- `telemetry/types.py` is the TypeScript SDK's `TelemetryEvent` field for field, including what it omits: `JSON.stringify` drops an `undefined`, so `payload()` drops a `None` and builds the keys in that SDK's declaration order. Both SDKs' events aggregate in one collector, so a field on one side only would split the aggregate it feeds; the next change here is the migration that takes the spec's telemetry schema out of Draft (PR 162, Q-10).
- Source information rides on `sdk_version` as `learning-commons-evaluators-python-<version>` — the identifier 0.2.0's `EvaluatorMetadata.sdk_version` already recorded. TypeScript sends a bare version there, so the prefix is what tells the collector which SDK sent an event, and no field exists on one side only. Attribution is the spec's: anonymous against a client id in `~/.config/learning-commons/config.json` — the file and the `telemetry.clientId` key the TypeScript SDK writes, so a host running both SDKs is one client — or identified when `telemetry.learning_commons_api_key` is set, sent as `X-API-Key`.
- Fire-and-forget means a background daemon thread, not a task on the caller's event loop: an event queued by `evaluate_sync` has to outlive the loop `asyncio.run` closes underneath it. Sending never blocks, never raises, and never changes a result; requests are bounded by the same 5s timeout TypeScript aborts at, a full queue drops events rather than memory, and an `atexit` flush keeps a short script from exiting before its last event leaves. `telemetry=False` builds no client at all, so nothing is constructed, queued, or started.
- No raw-input option, per the 2026-09-14 scope change: `record_raw_inputs` was dropped from `TelemetryOptions` in #278 as `recordInputs` was from TypeScript in #274. An event carries a text length, never the text, and there is no field it could be put in.

Verification, as the ticket asks, is against the TypeScript events as the oracle rather than against restated expectations. `cross-sdk-events.test.ts` runs six cases — single-step success, rejected input, provider failure, model override, multi-step success, multi-step failure after one step — against a captured sink and records each case with the event it produced; `test_cross_sdk_events.py` replays the same cases and compares field for field, field order included, with only the clock, wall-clock latency, and `sdk_version` masked. Regenerating the file (`UPDATE_TELEMETRY_ORACLE=1 npm run test:unit`) is what committing both SDKs to a format change looks like. Alongside it, `test_emission.py` runs over the registry, so a later phase cannot add an evaluator that stays silent, and asserts no event carries any free text an evaluator was given.

`tests/conftest.py` routes every test's telemetry into an in-process sink, so the suite cannot post to the collector; that sink is also what the telemetry tests read.

Verification: 875 Python tests and 1,390 TypeScript unit tests pass; Python coverage 98%, the new modules at 99-100%; ruff, ruff format, and mypy clean.

**Thanks for your interest in our project\!**

This repository is currently being maintained internally, so we’re not accepting external contributors.

However, if you’ve discovered a bug or have a feature suggestion, please feel free to open an issue or reach out to us at [support@learningcommons.org](mailto:support@learningcommons.org). 

If you believe you have found a security issue, please responsibly disclose by contacting us at [security@learningcommons.org](mailto:security@learningcommons.org).

If you are interested in partnering with us, fill out [this form](https://learningcommons.org/contact/?utm_source=github&utm_medium=overview&utm_campaign=privatebeta) to get in touch.


sdks/python/src/learning_commons_evaluators/evaluators/single_step.py
```diff
@@ -134,79 +133,80 @@ async def evaluate(
...
-            self.logger.info(
-                "Starting %s evaluation",
-                self.metadata.label,
-                extra={**context, "grade_level": grade_level, "text_length": len(text)},
-            )
+        with self._telemetry_run(self.provider.label) as run:
+            try:
+                values = validate_inputs(raw, self.contract.input_schema)
+                text = values.get(self._text_field, "") if self._text_field else ""
+                run.text_length = len(text)
```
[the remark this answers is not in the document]
michael-berger-czi: Comment from Claude: Fixed in 82438b3. Confirmed: this SDK reported Python `len()` (code points) where the TypeScript oracle reports `String.length` (UTF-16 code units), so one emoji was 1 here and 2 there — into a collector that sums both SDKs.

Worth adding that the cross-SDK suite could not have caught it: `sdks/typescript/tests/fixtures/telemetry-events.json` contains zero non-ASCII characters, so the "compared field for field" promise in that test's own docstring held only for ASCII input.

`telemetry/utils.py` now has `utf16_length`, used at both `run.text_length` sites, with six cases pinning the count across empty, ASCII, BMP (`café`, where it agrees with `len`) and non-BMP input.

One gap left open deliberately: the guard is Python-side only. Nothing yet asserts that TypeScript *agrees* — closing that needs a non-BMP case in `cross-sdk-events.test.ts` plus a regenerated oracle, which is a TypeScript change and out of scope for this PR.

sdks/python/src/learning_commons_evaluators/evaluators/multi_step.py
```diff
@@ -184,97 +183,96 @@ async def evaluate(
...
-                self.metadata.label,
-                extra={**context, "grade_level": grade_level, "text_length": len(text)},
-            )
+        # Until a step starts, the model named is the one the first declared step would
+        # call, so a failure before any of them — a rejected input — still reports a model.
+        with self._telemetry_run(self._providers[self._steps[0].id].label) as run:
+            try:
+                values = validate_inputs(raw, self.contract.input_schema)
+                text = values.get(self._text_field, "") if self._text_field else ""
+                run.text_length = len(text)
```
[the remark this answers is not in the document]
michael-berger-czi: Comment from Claude: Fixed in 82438b3, same change as the single-step path — `run.text_length` here now uses `utf16_length` too, so both bases report the count the TypeScript events report.

sdks/python/src/learning_commons_evaluators/telemetry/utils.py
```diff
@@ -0,0 +1,100 @@
...
+    A filesystem that cannot be read or written (a read-only container, a serverless
+    sandbox) is not a failure: the id is then per-process, and telemetry stays anonymous
+    either way.
+    """
+    global _client_id
+    if _client_id is not None:
+        return _client_id
+
+    path = config_file()
+    stored = _read_config(path)
```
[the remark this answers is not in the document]
michael-berger-czi: Comment from Claude: Valid, but not fixing it here — the same exposure is in the TypeScript SDK, so a Python-only patch would half-fix a shared file.

Confirmed on both halves of the finding:

- **In-process race.** `client_id()` is an unsynchronized check-then-act on a module global, and it is reached from `BaseEvaluator.__init__`, so a threaded service constructing evaluators per request can race on first use. Consequence is bounded: a few events attributed to a second anonymous ID before the file settles.
- **Non-atomic write**, which is the more substantive part. `_write_config` uses `path.write_text`, and `sdks/typescript/src/telemetry/utils.ts:48` uses a plain `writeFileSync` on the *same* file. Concurrent processes can tear it, and then `_read_config` swallows the invalid JSON and starts fresh — so the persistent client ID silently churns, and the torn write drops keys the other SDK owns, which contradicts the promise in this function's own docstring that "a key this SDK does not know about is not ours to drop."

So the fix wants to be both SDKs at once: serialize initialization, and write temp-then-rename. Deferred rather than done piecemeal; the failure mode is telemetry attribution, not evaluation correctness.

sdks/python/src/learning_commons_evaluators/evaluators/base.py
```diff
@@ -174,6 +203,58 @@ def _create_configured_provider(self, default: Provider, default_model: str) ->
...
+        """
+        run = TelemetryRun(provider=provider)
+        try:
+            yield run
+        except Exception as error:
+            self._emit(run, status="error", error_code=type(error).__name__)
+            raise
+        self._emit(run, status="success")
+
+    def _emit(
```
adnanrhussain, line 233: P1 - Pls wrap the body in try / except. logging at warning. Nothing in the current construction can realistically raise, so this is defence-in-depth

sdks/python/src/learning_commons_evaluators/evaluators/multi_step.py
```diff
@@ -184,97 +184,96 @@ async def evaluate(
...
+
+                plan = self._plan(values)
+                outputs: dict[str, Any] = {}
+                computed: dict[str, str] = {}
+
+                for step in plan:
+                    provider = self._providers[step.id]
+                    # Named before the call rather than after it, so a step that fails is
+                    # reported against its own model and not against the last one to finish.
+                    run.provider = provider.label
```
adnanrhussain, line 211: P1 - I think this is overwritten on every step, so maybe make run.provider the +- joined label of the distinct models that ran, in run order

sdks/python/src/learning_commons_evaluators/telemetry/client.py
```diff
@@ -0,0 +1,233 @@
...
+            thread.join(timeout)
+
+    # --- the sender's side -------------------------------------------------------------
+
+    def _start(self) -> None:
+        with self._starting:
+            if self._thread is not None and self._thread.is_alive():
+                return
+            self._thread = threading.Thread(
+                target=self._run, name="learning-commons-telemetry", daemon=True
```
adnanrhussain, line 132: P1 - I ran mutation tests, and flipping the daemon to False, does not fail any test. Can we assert the flag directly - assert sender()._thread.daemon is True after a submit

sdks/python/src/learning_commons_evaluators/telemetry/utils.py
```diff
@@ -0,0 +1,115 @@
...
+    return f"{SDK_IDENTIFIER}-{__version__}"
+
+
+def utc_timestamp() -> str:
+    """Now, as the event carries it: ``2026-09-17T18:04:05.123Z``.
+
+    Millisecond precision and a ``Z`` suffix, which is what JavaScript's
+    ``Date.toISOString`` produces and therefore what the collector already parses.
+    """
+    now = datetime.now(timezone.utc)
```
adnanrhussain, line 41: P2 - It looks like the tests only assert the shape. But if add this assertion we risk flaky tests. Perhaps we can freeze the clock and assert the frozen value?

sdks/python/src/learning_commons_evaluators/telemetry/run.py
```diff
@@ -0,0 +1,74 @@
...
+
+    #: ``provider:model`` for the model answering now: the last step to start, so a failure
+    #: reports the step that was running rather than one that already finished. Before any
+    #: step runs — a validation failure, say — it is the model that would have been called.
+    provider: str
+    #: Length of the evaluator's primary text input; ``0`` until the inputs are validated.
+    #: Counted in characters, where JavaScript counts UTF-16 code units, so text outside the
+    #: BMP reads one lower per astral character than the same text does from the TypeScript
+    #: SDK. The field is a magnitude in an aggregate, not an identity, so it is counted the
+    #: way each language counts.
```
adnanrhussain, line 33: Based on the copilot finding and your update, this may now be invalid

sdks/python/src/learning_commons_evaluators/telemetry/utils.py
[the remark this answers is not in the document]
adnanrhussain: pls add a TODO for now

adnanrhussain (approved):
LGTM

sdks/python/src/learning_commons_evaluators/telemetry/run.py
michael-berger-czi: Comment from Claude: Right, and it was stale from the moment the UTF-16 fix landed — I changed the counting and left the prose describing the old behaviour. It said:

> Counted in characters, where JavaScript counts UTF-16 code units, so text outside the BMP reads one lower per astral character than the same text does from the TypeScript SDK.

That now tells the reader the two SDKs disagree when they agree. Rewritten in `921b83f` to say the count is UTF-16 code units in both, with a pointer to `utf16_length`.

sdks/python/src/learning_commons_evaluators/evaluators/base.py
michael-berger-czi: Comment from Claude: Done in `921b83f`, and I think it is stronger than defence-in-depth: spec §7 requires that logging never "throw, block, or alter results", and telemetry sits on the same path. As written, a raise inside `_emit` propagated out of `evaluate()` and failed an evaluation that had already succeeded.

`_emit` now wraps the send and logs at warning; the body moved to `_send_event` so the guard is one obvious line rather than an indented block. Added a test that patches `TelemetryClient.send` to raise and asserts the evaluation still returns its result with only a warning logged — otherwise the guard is exactly the kind of code that rots untested.

sdks/python/src/learning_commons_evaluators/telemetry/client.py
michael-berger-czi: Comment from Claude: Good catch, and confirmed — zero occurrences of `daemon` anywhere under `tests/` before this. Asserted directly in `921b83f`, on the module-level sender as you suggested:

```python
make_client().send(EVENT)
thread = telemetry_client.sender()._thread
assert thread is not None and thread.daemon is True
```

I re-ran your mutation to check the assertion actually bites: flipping `daemon=True` to `False` now fails this test and nothing else, which is the shape you would want.

sdks/python/src/learning_commons_evaluators/telemetry/utils.py
michael-berger-czi: Comment from Claude: Agreed, and freezing the clock avoids the flakiness rather than risking it. Added in `921b83f` alongside the shape check, monkeypatching `utils.datetime` to a fixed instant.

Chose `123_789` microseconds deliberately: it pins that milliseconds are **truncated** rather than rounded (`.123`, not `.124`), which is the behaviour that could silently drift and which the regex could never catch. The `Z` suffix is pinned by the same assertion.

sdks/python/src/learning_commons_evaluators/telemetry/utils.py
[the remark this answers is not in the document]
michael-berger-czi: Comment from Claude: Added in `921b83f` — a TODO at `client_id()` naming both halves (unsynchronized init, non-atomic write), why it is deferred rather than fixed here (the TypeScript SDK writes the same shared file the same way), and DSCR-2215 as the epic it belongs to. So the deferral is visible where someone will hit it, not just in this thread.

sdks/python/src/learning_commons_evaluators/evaluators/multi_step.py
michael-berger-czi: Comment from Claude: Confirmed, and agreed it is inconsistent — `run.provider` ends up as the last step's label, while the envelope reports `"+".join(...)` of the distinct models in run order. So one multi-step evaluation currently answers "which model ran" two different ways.

Not changing it in this PR, for two reasons.

**It is a shared-format change, not a Python one.** The cross-SDK oracle records a single label for the multi-step cases:

```
multi-step/success: provider='openai:gpt-4o-2024-08-06'
```

and that fixture is generated from TypeScript. Making Python emit the joined label breaks the parity suite until TypeScript emits it too — which is exactly what that suite is for.

**The current behaviour is deliberate for error events.** `run.py` documents it as "the last step to start, so a failure reports the step that was running rather than one that already finished", which is the more useful attribution when a multi-provider run fails partway. A blanket join would lose that. My guess at the right rule is joined-so-far — on success every model that ran, on error the steps that ran plus the one that failed — but that is a semantics decision for the event format rather than something to slip into this PR.

Happy to file it against DSCR-2215 with the constraint written down, unless you would rather it rode along with a TypeScript change you already have in flight.

sdks/python/src/learning_commons_evaluators/telemetry/utils.py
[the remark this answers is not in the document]
michael-berger-czi: Comment from Claude: Filed as **DSCR-2250** and repointed the marker (`04d2974`) — the TODO had said `DSCR-2215`, which is the follow-ups epic rather than an issue, so it pointed somewhere a reader could not actually find.

The ticket records both halves and why neither is safe to fix in one SDK alone:

- **Unsynchronized init** — check-then-act on a module global, reached from `BaseEvaluator.__init__`, so a threaded service can race on first use. Bounded: a few events under a second anonymous id.
- **Non-atomic write** — the substantive one. `_write_config` uses `path.write_text` and `sdks/typescript/src/telemetry/utils.ts:48` uses a plain `writeFileSync` on the *same* `~/.config/learning-commons/config.json`. A torn file makes `_read_config` start fresh, so the id that exists to be persistent silently churns, and the write drops keys the other SDK owns — against the promise in that function's own docstring.

Removing the TODO is an acceptance criterion on the ticket, so the marker and the fix cannot drift apart.

sdks/python/src/learning_commons_evaluators/evaluators/multi_step.py
michael-berger-czi: Comment from Claude: Filed as **DSCR-2251**, together with a second change to the same event — carrying the evaluator's `stable_id`, which was requested separately and hits the identical constraint.

Grouping them because the expensive part is shared: agreeing the semantics, changing both SDKs, regenerating the oracle, and confirming the collector accepts the change. Two separate passes over the shared event format would pay that twice.

The ticket records your point as I understood it, including the tension I flagged above: the rule that keeps both properties is probably joined-so-far — on success every model that ran, on error the steps that ran plus the one that failed — so the error case keeps the attribution `run.py` deliberately gives it. Flagged as a semantics decision rather than an implementation detail, so it gets decided rather than assumed.

sdks/python/src/learning_commons_evaluators/evaluators/base.py
```diff
@@ -58,6 +70,23 @@ def __init__(self, config: EvaluatorConfig | None = None, /, **fields: Any) -> N
         self._override_provider = self._validate_model_override(config.model_override)
         self._validate_credentials()
 
+        telemetry = config.telemetry_options
+        # Disabled means disabled: no client, so nothing is built, queued, or sent, and the
+        # process never starts a sender thread on this evaluator's account.
```
czi-fsisenda: Is all of this from config / constants? If that's the case, creating the telemetry client could be refactored out into a helper to keep this file simpler.

czi-fsisenda: send_event could live in a helper too.

sdks/python/src/learning_commons_evaluators/telemetry/client.py
```diff
@@ -0,0 +1,233 @@
...
+
+#: The collector both SDKs post to.
+COLLECTOR_ENDPOINT = "https://api.learningcommons.org/evaluators-telemetry/v1/events"
+
+#: How long one request may take, matching the TypeScript SDK's abort signal.
+TIMEOUT_SECONDS = 5.0
+
+#: Events waiting to be sent. A caller evaluating faster than the collector accepts loses
+#: events rather than memory; telemetry is a sample, not a ledger.
+QUEUE_CAPACITY = 100
```
czi-fsisenda: These could be in config so we can use different values without changing code

sdks/python/src/learning_commons_evaluators/telemetry/client.py
```diff
@@ -0,0 +1,233 @@
...
+
+
+@dataclass(frozen=True)
+class TelemetryClientConfig:
+    """What the client needs to send an event, assembled once per evaluator."""
+
+    endpoint: str
```
czi-fsisenda: This could default to COLLECTOR_ENDPOINT?

